PR #1142, opened 2014-11-06
Support for Subversion

fladi:
This adds a new VCS module called `svn.py` that uses [PySVN](http://pysvn.tigris.org) to gather information if the current working directory is part of a Subversion repository.

As there is no clear definition of a branch in Subversion, this module uses the difference between the actual (remote) repository root URL and the URL of the checked out copy as the branch name. If the Subversion repository follows the usual conventions, this would lead to names like `/trunk`, `/branch/feature/foo` or `/tag/1.0.1`. The currently checked our revision number is displayed after the branch name (as in `/trunk:r1234`). If the local copy was checked out from the remote repository root URL, the branch name will simply be `/`.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -19,6 +19,8 @@ Generic requirements
   repositories.
 * ``bzr`` python package (note: *not* standalone executable). Required to work 
   with bazaar repositories.
+* ``pysvn`` python package (note: *not* ``svn`` standalone executable). Required 
+  to work with subversion repositories.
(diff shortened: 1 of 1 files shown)


ZyX-I:
@fladi If you know how to build pysvn you can post a script here so that it can be used with https://github.com/powerline/bot-ci (do not run PR there (more precise: if you run the PR do not expect it to build): uploading build results requires deployment key, deployment key requires encryption and encrypted environment variables are not available in PRs thus PRs builds are disabled there because they should fail otherwise).

PR #1200, opened 2014-11-29
Obtain tmux colors and dividers from the configuration

ZyX-I:
No support for `--theme_option` or `--config`: I think it is needed to use some variable _other_ then `POWERLINE_COMMAND` to specify configuration overrides both to `powerline-config` and `powerline`. Possibilities:
1. Share code in `powerline/commands/*.py` and call both `POWERLINE_CONFIG_COMMAND` and `POWERLINE_COMMAND` like `$POWERLINE_…COMMAND $POWERLINE_COMMAND_ARGS`.
2. Use environment variables like those already used in zsh/zpython to specify configuration.
3. Use `$POWERLINE_COMMAND --setup` to set up tmux, not `$POWERLINE_CONFIG_COMMAND tmux setenv`.

### DIFF
--- a/powerline/bindings/config.py
+++ b/powerline/bindings/config.py
@@ -142,6 +142,10 @@ def init_environment(pl, args):
 		else:
 			set_tmux_environment(varname, 'colour' + str(get_highlighting(group)[attr][0]))
 
+	left_dividers = powerline.renderer.theme.dividers['left']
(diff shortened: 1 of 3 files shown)


phongvcao:
Today I am back to using Tmux and Powerline and the Black spot still persists. Does anyone know how to setup Solarized Powerline for Tmux so that we can get rid of this little black spot on the left of the bottom Tmux powerline bar?

selection_002

PR #1256, opened 2015-01-07
[WIP] Add BAR support (Bar ain't recursive)

S0lll0s:
Tiny renderer + i3 bindings. Don't merge yet, have some (hopefully) minor fixing left to do

### DIFF
--- a/docs/source/usage/wm-widgets.rst
+++ b/docs/source/usage/wm-widgets.rst
@@ -46,8 +46,8 @@ Add the following to your :file:`~/.config/qtile/config.py`:
 
 .. _bar-usage:
 
-bar
-===
+LemonBoy’s bar
+=============
 
 .. note:: There is currently only a script to run inside i3, but you can just strip the
           i3-specific parts of the binding.
@@ -55,11 +55,11 @@ bar
 To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
 options, for example like this::
 
-    python powerline-bar.py | bar -f "-xos4-*-*-*-*-*-*-*-*-*-*-*-*-*
+    python powerline-bar.py | bar -f "-xos4-*"
 
 to run with i3, simply ``exec`` this in your i3 config file::
 
-    exec python powerline-bar.py | bar -f "-xos4-*-*-*-*-*-*-*-*-*-*-*-*-*
+    exec python powerline-bar.py | bar -f "-xos4-*"
 
 See the `bar documentation <https://github.com/LemonBoy/bar>`_ for more information and options.
 
--- a/docs/source/usage/wm-widgets.rst
+++ b/docs/source/usage/wm-widgets.rst
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
        ),
    ]
 
+.. _bar-usage:
+
+bar
+===
+
+.. note:: There is currently only a script to run inside i3, but you can just strip the
+          i3-specific parts of the binding.
+
+To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
+options, for example like this::
+
+    python powerline-bar.py | bar -f "-xos4-*-*-*-*-*-*-*-*-*-*-*-*-*
+
+to run with i3, simply ``exec`` this in your i3 config file::
+
+    exec python powerline-bar.py | bar -f "-xos4-*-*-*-*-*-*-*-*-*-*-*-*-*
+
+See the `bar documentation <https://github.com/LemonBoy/bar>`_ for more information and options.
+
 I3 bar
 ======
 
-.. note:: Until the patch is done in i3, you will need a custom ``i3bar`` build
-          called ``i3bgbar``. The source is available `here 
-          <https://github.com/S0lll0s/i3bgbar>`_.
+.. note:: As the patch to include background-colors in i3bar is likely not to be merged,
+          it is recommended to instead run ``bar`` (see above).
+          The source for i3bgbar is however still available `here <https://github.com/S0lll0s/i3bgbar>`_.
 
 Add the following to your :file:`~/.i3/config`::
 
--- a/powerline/bindings/i3/bar.py
+++ b/powerline/bindings/i3/bar.py
@@ -29,7 +29,7 @@ if __name__ == '__main__':
 		with lock:
 			ln = u'%{l}'
 			ln += powerline.render(side='left')
-			# ln += u'%{r}'
+			ln += u'%{r}'
 			ln += powerline.render(side='right')
 			print( ln.encode('utf-8') )
 			sys.stdout.flush()
@@ -41,5 +41,4 @@ if __name__ == '__main__':
 			render()
 			time.sleep(max(interval - (monotonic() - start_time), 0.1))
 	except KeyboardInterrupt, e:
-		print( "caught" )
 		exit()
--- a/powerline/bindings/i3/bar.py
+++ b/powerline/bindings/i3/bar.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+import sys
+import time
+
+from threading import Lock
+
+import i3
+
+from powerline import Powerline
+from powerline.lib.monotonic import monotonic
+
+if __name__ == '__main__':
+	name = 'wm'
+	if len(sys.argv) > 1:
+		name = sys.argv[1]
+
+	powerline = Powerline( name, renderer_module='bar' )
+	powerline.update_renderer()
+
+	interval = 0.5
+
+	lock = Lock()
+
+	def render(event=None, data=None, sub=None):
+		global lock
+		with lock:
+			ln = u'%{l}'
+			ln += powerline.render(side='left')
+			# ln += u'%{r}'
+			ln += powerline.render(side='right')
+			print( ln.encode('utf-8') )
+			sys.stdout.flush()
+
+	sub = i3.Subscription(render, 'workspace')
+	try:
+		while True:
+			start_time = monotonic()
+			render()
+			time.sleep(max(interval - (monotonic() - start_time), 0.1))
+	except KeyboardInterrupt, e:
+		print( "caught" )
+		exit()
--- a/powerline/bindings/i3/powerline-bar.py
+++ b/powerline/bindings/i3/powerline-bar.py
@@ -4,11 +4,8 @@ from __future__ import (unicode_literals, division, absolute_import, print_funct
 
 import sys
 import time
-
-from threading import Lock
-
 import i3
-
+from threading import Lock
 from powerline import Powerline
 from powerline.lib.monotonic import monotonic
 
--- a/powerline/bindings/i3/powerline-bar.py
+++ b/powerline/bindings/i3/powerline-bar.py
@@ -21,24 +21,20 @@ if __name__ == '__main__':
 	powerline.update_renderer()
 
 	interval = 0.5
-
 	lock = Lock()
 
 	def render(event=None, data=None, sub=None):
 		global lock
 		with lock:
-			ln = u'%{l}'
+			ln = '%{l}'
 			ln += powerline.render(side='left')
-			ln += u'%{r}'
+			ln += '%{r}'
 			ln += powerline.render(side='right')
 			print( ln.encode('utf-8') )
 			sys.stdout.flush()
 
 	sub = i3.Subscription(render, 'workspace')
-	try:
-		while True:
-			start_time = monotonic()
-			render()
-			time.sleep(max(interval - (monotonic() - start_time), 0.1))
-	except KeyboardInterrupt, e:
-		exit()
+	while True:
+		start_time = monotonic()
+		render()
+		time.sleep(max(interval - (monotonic() - start_time), 0.1))
--- a/powerline/bindings/i3/powerline-bar.py
+++ b/powerline/bindings/i3/powerline-bar.py
@@ -20,6 +20,11 @@ if __name__ == '__main__':
 	interval = 0.5
 	lock = Lock()
 
+	def encode( str ): return str.encode('utf-8')
+
+	if sys.version_info > (3,0):
+		def encode( str ): return str
+
 	def render(event=None, data=None, sub=None):
 		global lock
 		with lock:
@@ -27,7 +32,7 @@ if __name__ == '__main__':
 			ln += powerline.render(side='left')
 			ln += '%{r}'
 			ln += powerline.render(side='right')
-			print( ln.encode('utf-8') )
+			print( encode(ln) )
 			sys.stdout.flush()
 
 	sub = i3.Subscription(render, 'workspace')
--- a/powerline/renderers/bar.py
+++ b/powerline/renderers/bar.py
@@ -13,8 +13,7 @@ class BarRenderer(Renderer):
 	'''
 
 	character_translations = Renderer.character_translations.copy()
-	character_translations[ord('{')] = '{{'
-	character_translations[ord('}')] = '}}'
+	character_translations[ord('%')] = '%%'
 
 	@staticmethod
 	def hlstyle(*args, **kwargs):
--- a/powerline/renderers/bar.py
+++ b/powerline/renderers/bar.py
@@ -2,6 +2,7 @@
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
 from powerline.renderer import Renderer
+from powerline.colorscheme import ATTR_BOLD, ATTR_ITALIC, ATTR_UNDERLINE
 
 
 class BarRenderer(Renderer):
@@ -9,22 +10,29 @@ class BarRenderer(Renderer):
 	(https://github.com/LemonBoy/bar)
 	'''
 
+	character_translations = Renderer.character_translations.copy()
+	character_translations[ord('{')] = '{{'
+	character_translations[ord('}')] = '}}'
+
 	@staticmethod
 	def hlstyle(*args, **kwargs):
 		# We don’t need to explicitly reset attributes, so skip those calls
 		return ''
 
 	def hl(self, contents, fg=None, bg=None, attrs=None):
-		text = u''
+		text = ''
 
 		if fg is not None:
 			if fg is not False and fg[1] is not False:
-				text +=  u'%{{F#ff{0:06x}}}'.format(fg[1])
+				text +=  '%{{F#ff{0:06x}}}'.format(fg[1])
 		if bg is not None:
 			if bg is not False and bg[1] is not False:
-				text +=  u'%{{B#ff{0:06x}}}'.format(bg[1])
+				text +=  '%{{B#ff{0:06x}}}'.format(bg[1])
 
-		return text + contents
+		if attrs & ATTR_UNDERLINE:
+			text += '%{+u}'
+
+		return text + contents + '%{F-B--u}'
 
 
 renderer = BarRenderer
--- a/powerline/renderers/bar.py
+++ b/powerline/renderers/bar.py
@@ -1,0 +1,30 @@
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+from powerline.renderer import Renderer
+
+
+class BarRenderer(Renderer):
+	'''BAR (Bar ain't recursive) rendere
+	(https://github.com/LemonBoy/bar)
+	'''
+
+	@staticmethod
+	def hlstyle(*args, **kwargs):
+		# We don’t need to explicitly reset attributes, so skip those calls
+		return ''
+
+	def hl(self, contents, fg=None, bg=None, attrs=None):
+		text = u''
+
+		if fg is not None:
+			if fg is not False and fg[1] is not False:
+				text +=  u'%{{F#ff{0:06x}}}'.format(fg[1])
+		if bg is not None:
+			if bg is not False and bg[1] is not False:
+				text +=  u'%{{B#ff{0:06x}}}'.format(bg[1])
+
+		return text + contents
+
+
+renderer = BarRenderer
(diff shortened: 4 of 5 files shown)


powerline/renderers/bar.py
```diff
@@ -0,0 +1,30 @@
...
+	(https://github.com/LemonBoy/bar)
+	'''
+
+	@staticmethod
+	def hlstyle(*args, **kwargs):
+		# We don’t need to explicitly reset attributes, so skip those calls
+		return ''
+
+	def hl(self, contents, fg=None, bg=None, attrs=None):
+		text = u''
```
ZyX-I: You must not use `u''`: there is `from __future__ import unicode_literals`.

powerline/renderers/bar.py
```diff
@@ -0,0 +1,30 @@
...
+
+		if fg is not None:
+			if fg is not False and fg[1] is not False:
+				text +=  u'%{{F#ff{0:06x}}}'.format(fg[1])
```
ZyX-I: Same.

powerline/renderers/bar.py
```diff
@@ -0,0 +1,30 @@
...
+		if bg is not None:
+			if bg is not False and bg[1] is not False:
+				text +=  u'%{{B#ff{0:06x}}}'.format(bg[1])
```
ZyX-I: Same.

powerline/bindings/i3/bar.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+import time
+
+from threading import Lock
+
+import i3
+
+from powerline import Powerline
+from powerline.lib.monotonic import monotonic
+
```
ZyX-I: Not enough blank lines.

powerline/bindings/i3/bar.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
```
ZyX-I: Bindings file should be named `powerline…`: e.g. `powerline-bar.py`. Note that this script is most likely be moved elsewhere and “elsewhere” does not necessary contain `powerline` in directory name.

powerline/bindings/i3/bar.py
```diff
@@ -0,0 +1,44 @@
...
+			sys.stdout.flush()
+
+	sub = i3.Subscription(render, 'workspace')
+	try:
+		while True:
+			start_time = monotonic()
+			render()
+			time.sleep(max(interval - (monotonic() - start_time), 0.1))
+	except KeyboardInterrupt, e:
+		exit()
```
ZyX-I: What is this doing here? Bindings will normally exit on KeyboardInterrupt and I do not see much sense in forcing them exit with zero error code. You definitely must not assign exception to `e` unless you are going to use it.

S0lll0s: Because of the lock the code takes ages to exit, I kindof fixed it like this (at least I think I did).

Either way this could easily be considered debug stuff so I'll remove it tomorrow

powerline/renderers/bar.py
S0lll0s: alright, didn't know, will fix

S0lll0s:
Works for me now, but I'd wait for confirmation from another actual system (this was VM and no extensive use). Check formatting again?

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+import time
+
+from threading import Lock
+
+import i3
+
+from powerline import Powerline
+from powerline.lib.monotonic import monotonic
+
```
ZyX-I: Not enough empty lines.

powerline/renderers/bar.py
```diff
@@ -0,0 +1,38 @@
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+from powerline.renderer import Renderer
+from powerline.colorscheme import ATTR_BOLD, ATTR_ITALIC, ATTR_UNDERLINE
+
+
+class BarRenderer(Renderer):
+	'''BAR (Bar ain't recursive) rendere
```
ZyX-I: “render**e**”?

powerline/renderers/bar.py
```diff
@@ -0,0 +1,38 @@
...
+	(https://github.com/LemonBoy/bar)
```
ZyX-I: I guess it should either be removed completely or put into description (+1 empty line and it must contain some text in addition to the link). It is also highly suggested not to have link as text, but use link with some caption like just `LemonBoy/bar`.

ZyX-I:
Also missing documentation update (docs/source/usage/wm-widgets.rst).

ZyX-I:
And tests in tests/test_configuration.py. It is not good that renderers (except shell ones and Vim) are almost untested, especially renderers which do not have functional tests.

S0lll0s:
I'm not sure how I would add the test case (I'm not experienced with unit tests, especially in python), maybe someone else could add it if I give the sample output?

ZyX-I:
@S0lll0s Tests there are actually not unit, they look more like functional. There is nothing special, just take existing test class (`TestVim`), existing theme (`config = { … 'themes/shell/default': {…} …}`), existing extension configuration (`config={ 'config': {… 'ext': {… 'vim': {…} …} …}}` and copy that, replacing various classes and identifiers with your names.

Specifically you need `theme/wm/default` (just make it a direct copy of `themes/shell/default`), `TestBAR` in place of `TestVim`, only one function `test_bar` and `from powerline import Powerline`. If I have not missed something then the result should look something like

```Patch
diff --git a/tests/test_configuration.py b/tests/test_configuration.py
index efc2147..6e14ff0 100644
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -39,6 +39,10 @@ config = {
 				'theme': 'default',
 				'colorscheme': 'default',
 			},
+			'wm': {
+				'theme': 'default',
+				'colorscheme': 'default',
+			},
 			'shell': {
 				'theme': 'default',
 				'colorscheme': 'default',
@@ -142,6 +146,14 @@ config = {
 			],
 		},
 	},
+	'themes/wm/default': {
+		'default_module': 'powerline.segments.common',
+		'segments': {
+			'left': [
+				highlighted_string('s', 'g1', width='auto'),
+			],
+		},
+	},
 }
 
 
@@ -800,6 +812,14 @@ class TestVim(TestCase):
 		sys.path.pop(0)
 
 
+class TestBar(TestCase):
+	def test_bar(self):
+		import powerline as powerline_module
+		with swap_attributes(config, powerline_module):
+			with get_powerline_raw(config, powerline_module.Powerline, ext='wm', renderer_module='bar') as powerline:
+				self.assertEqual(powerline.render(side='left'), '{SOME STRING HERE}')
+
+
 if __name__ == '__main__':
 	from tests import main
 	main()
```

ZyX-I:
By the way, you need to rebase. Also I have removed all references to “you” in documentation except for some certain places (tips-and-tricks, troubleshooting, guide). And you are importing unused `ATTR_*` as far as I see.

docs/source/usage/wm-widgets.rst
```diff
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
        ),
    ]
 
+.. _bar-usage:
+
+bar
```
ZyX-I: This should be capital.

ZyX-I:
> Also I have removed all references to “you” in documentation except for some certain places (tips-and-tricks, troubleshooting, guide).

Which means that merge conflict appears in documentation. Just use “ours” merging strategy option when rebasing.

docs/source/usage/wm-widgets.rst
S0lll0s: I'm still not sure how to put it, the original bar is written all-lowercase (like unix commands use to) but it's hard to tell from the noun then.

ZyX-I: @S0lll0s If you do not want to make bar uppercase write author name there. It is not the only package named “bar” in any case: I have `sys-apps/bar` from http://clpbar.sourceforge.net/ which is “Console Progress Bar” in main portage tree.

S0lll0s: I know. So you suggest I name the section LemonBoy/bar?

ZyX-I: @S0lll0s Of course, *not*. This is section name, not an identifier. Something like `LemonBoy’s bar` (if you use this better copy from here: I use U+2019 for apostrophe all over documentation if I write it myself. This variant usually looks better then `LemonBoy's bar` with U+0027).

docs/source/usage/wm-widgets.rst
```diff
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
...
+===
+
+.. note:: There is currently only a script to run inside i3, but you can just strip the
+          i3-specific parts of the binding.
+
+To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
+options, for example like this::
+
+    python powerline-bar.py | bar -f "-xos4-*-*-*-*-*-*-*-*-*-*-*-*-*
```
ZyX-I: Just mentioned this: this code is lacking trailing QUOTATION MARK. Same for the code below I guess.

docs/source/usage/wm-widgets.rst
ZyX-I: `Giuseppe’s bar` should be better, assuming it is his name and not some random placeholder string.

S0lll0s:
I'm trying to build the test case now, but powerline.render always returns

    Config file not found in search paths (/home/s0lll0s/powerline/powerline/config_files, ~/.config/powerline): config.json

instead of the rendered string, even though a config.json exists in both places.

ZyX-I:
@S0lll0s It must work regardless of whether `config.json` is found in these paths because paths are ignored. This test overrides things that load configuration from files with things that load configuration from `config` dictionary provided to `swap_attributes`.

ZyX-I:
> I'm trying to build the test case now, but powerline.render always returns

With the code in the last commit you must not have `powerline.render` at all, it is strange that it returns anything. `get_powerline_raw` should not receive `powerline_module`.

docs/source/usage/wm-widgets.rst
```diff
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
        ),
    ]
 
+.. _bar-usage:
+
+LemonBoy’s bar
+=============
```
ZyX-I: This bar is too short. ReST is not markdown, *its parser explicitly checks whether underline is long enough* and does not recognize too short bars as a headers.

tests/test_configuration.py
```diff
@@ -799,6 +824,24 @@ def setUpClass(cls):
 	def tearDownClass(cls):
 		sys.path.pop(0)
 
+class TestBar(TestCase):
+	def test_bar(self):
+		import powerline as powerline_module
+		with swap_attributes(config, powerline_module):
+			with get_powerline_raw(config, powerline_module, ext='wm', renderer_module='bar') as powerline:
+				self.assertEqual("%{{l}}{}%{{r}}{}".format(powerline.render(side='left'), powerline.render(side='right')), "SOME_STRING_HERE")
```
ZyX-I: If everything is right `powerline.render` is being taked from `bar` module. `powerline.render` in bar module does not support `side` argument, so this code should only show a `TypeError`.

tests/test_configuration.py
```diff
@@ -799,6 +824,24 @@ def setUpClass(cls):
...
+	def test_bar_escape(self):
+		from powerline.shell import ShellPowerline
+		import powerline as powerline_module
+		config['themes/wm/default']['segments']['left'] = {
+			highlighted_string('%{asd}', 'hl1'),
+			highlighted_string('10% %', 'hl2'),
+		}
+		with swap_attributes(config, powerline_module):
+			with get_powerline_raw(config, powerline_module, ext='wm', renderer_module='bar') as powerline:
+				self.assertEqual(powerline.render(side='left'), "SOME_STRING_HERE")
```
ZyX-I: Same here.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+		def encode( str ): return str
+
+	def render(event=None, data=None, sub=None):
+		global lock
+		with lock:
+			ln = '%{l}'
+			ln += powerline.render(side='left')
+			ln += '%{r}'
+			ln += powerline.render(side='right')
+			print( encode(ln) )
```
ZyX-I: There is a space after `(` and before `)`. It should not be there.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+	if len(sys.argv) > 1:
+		name = sys.argv[1]
+
+	powerline = Powerline( name, renderer_module='bar' )
+	powerline.update_renderer()
+
+	interval = 0.5
+	lock = Lock()
+
+	def encode( str ): return str.encode('utf-8')
```
ZyX-I: No spaces before/after `(` and `)`, do not use single-line functions. Do not use this `encode` functions at all, there is `string` function in `powerline.lib.unicode`.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+from threading import Lock
+from powerline import Powerline
+from powerline.lib.monotonic import monotonic
+
+if __name__ == '__main__':
+	name = 'wm'
+	if len(sys.argv) > 1:
+		name = sys.argv[1]
+
+	powerline = Powerline( name, renderer_module='bar' )
```
ZyX-I: Spaces.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+import time
+import i3
+from threading import Lock
+from powerline import Powerline
+from powerline.lib.monotonic import monotonic
+
+if __name__ == '__main__':
+	name = 'wm'
+	if len(sys.argv) > 1:
+		name = sys.argv[1]
```
ZyX-I: This is `ext`, not `name`. Why do you need to override it at all?

powerline/bindings/i3/powerline-bar.py
ZyX-I: By the way, why do you need `encode` in first place? `powerline-i3.py` is just running `print`?

powerline/bindings/i3/powerline-bar.py
ZyX-I: I guess you just copied it from `powerline-i3.py`. Do not edit, if I get rid of this I will do this for both files at once.

powerline/renderers/bar.py
```diff
@@ -0,0 +1,39 @@
...
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+from powerline.renderer import Renderer
+from powerline.colorscheme import ATTR_BOLD, ATTR_ITALIC, ATTR_UNDERLINE
+
+
+class BarRenderer(Renderer):
+	'''bar (bar ain't recursive) renderer
+
+
```
ZyX-I: Too many newlines.

ZyX-I:
It appears that you are having problems because I forgot to instruct you about adding overrides (worse: I forgot to add overrides to `TestVim` second function which only worked because it follows `test_environ_update` which actually should not leave modified state). I will fix these problems, but you will need to rebase to get this fix.

ZyX-I:
With new fix (not yet committed) test will look like in my diff, but with additional keyword argument `replace_gcp=True` added to `get_powerline_raw`.

powerline/bindings/i3/powerline-bar.py
S0lll0s: because otherwise it didn't work in python3 (python2 required the str.encode(), python3 failed with that)

S0lll0s:
So now I can try my code with replace_gcp=True?

tests/test_configuration.py
S0lll0s: but in i3/powerline-bar.py it works with left/right

ZyX-I: @S0lll0s Sorry, I have mistaken the `render` method location. I mean, I treated `def render` in `powerline-bar.py` as `def render` located in `bar.py`, using class function. By the way, it is where this method should actually be.

ZyX-I: Ah, no. If you have some plans about disabling left or right parts it is better to leave everything as-is, *except* that you must not use `%{l}` and `%{r}` in tests. If you have no plans it is better to move it to `bar.py` like Vim renderer in `powerline/renderers/vim.py`.

powerline/bindings/i3/powerline-bar.py
ZyX-I: How it does not work? Python-2 prints both `str` and `unicode` instances just fine as long as you have correct locale. Python-2 and Python-3 *both* do not print unicode strings fine if you do not have correct locale and you have unicode string in Python-3.

ZyX-I:
> So now I can try my code with replace_gcp=True?

After a rebase, yes. And not your code, there are other issues I pointed.

tests/test_configuration.py
S0lll0s: as I am currently only piping the output to bar, I'll leave it as is (except for the test), should be a small change if there ever is a need to anyway

powerline/bindings/i3/powerline-bar.py
S0lll0s: For some reason it hex-encoded every byte, whether it was ASCII or not and somehow didn't work in general. Output was like this for example:

    \xc2\xa0\xe2\x8c\x9a\xc2\xa020:43%{F-B--u}%{F#ffd0d0d0}%{B#ff303030}\xc2\xa0\xee\x82\xb2

We tried a few things (I dont have py3 installed, friend of me had) and that was the only solution that worked for both py2 and 3

ZyX-I: @S0lll0s `powerline.render` normally returns a unicode string. You can’t print it with this result, it is not even `repr()`.

docs/source/usage/wm-widgets.rst
```diff
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
...
+
+.. note:: There is currently only a script to run inside i3, but you can just strip the
+          i3-specific parts of the binding.
```
ZyX-I: This means that you must have `python powerline-bar.py --i3` and must not import or use `i3` without this switch.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+import sys
+import time
```
ZyX-I: Not enough blank lines.

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+import i3
```
ZyX-I: Out-of-order import (see contributing, order is described there).

powerline/bindings/i3/powerline-bar.py
```diff
@@ -0,0 +1,42 @@
...
+from threading import Lock
```
ZyX-I: Not enough blank lines.

tests/test_configuration.py
```diff
@@ -799,6 +824,24 @@ def setUpClass(cls):
 	def tearDownClass(cls):
 		sys.path.pop(0)
```
ZyX-I: Not enough blank lines. I really suggest using syntastic or the like if you are not using Vim.

docs/source/usage/wm-widgets.rst
```diff
@@ -44,12 +44,31 @@ Add the following to your :file:`~/.config/qtile/config.py`:
...
+
+To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
+options, for example like this::
+
+    python powerline-bar.py | bar -f "-xos4-*"
```
ZyX-I: Where is `powerline-bar` located? Documentation must show the path.

ZyX-I: I also do not think that `-f` should be here.

tests/test_configuration.py
```diff
@@ -799,6 +824,24 @@ def setUpClass(cls):
...
+	def test_bar(self):
+		import powerline as powerline_module
+		with swap_attributes(config, powerline_module):
+			with get_powerline_raw(config, powerline_module, ext='wm', renderer_module='bar') as powerline:
+				self.assertEqual("%{{l}}{}%{{r}}{}".format(powerline.render(side='left'), powerline.render(side='right')), "SOME_STRING_HERE")
+
+	def test_bar_escape(self):
+		from powerline.shell import ShellPowerline
+		import powerline as powerline_module
+		config['themes/wm/default']['segments']['left'] = {
```
ZyX-I: Did you ever run your tests? WTF is set literal doing here?

ZyX-I:
Please check out #1285. I have all spotted issues fixed there.

tests/test_configuration.py
S0lll0s: I am using vim...

tests/test_configuration.py
S0lll0s: I couldn't, because of the config issue. Haven't rebased yet.

S0lll0s: And what set literal?

ZyX-I: https://docs.python.org/2.7/reference/expressions.html#set-displays. Set literal is `"{" expression_list "}"` part, a tuple or a list should work here. Set literals may actually also work… if only dictionaries were hashable and only by a chance because order is not guaranteed. They are not and you get `TypeError` when trying to construct set literal with the output of `highlighted_string`.

ZyX-I: And they would not work in Python-2.6 which powerline still supports.

PR #1271, opened 2015-01-11
Add support for cygwin battery status using cdll

aikeru:
In Cygwin on Windows, python doesn't have win32com (no COM support). Calling out to COM did not seem to be trivial under cygwin, but WMIC is easy.

I am currently using this fork locally and it's working great! :)

### DIFF
--- a/powerline/segments/common/bat.py
+++ b/powerline/segments/common/bat.py
@@ -148,7 +148,7 @@ def _get_battery(pl):
 			pl.debug('Using GetSystemPowerStatus')
 
 		return _get_capacity
-	
+
 	raise NotImplementedError
 
 
--- a/powerline/segments/common/bat.py
+++ b/powerline/segments/common/bat.py
@@ -95,6 +95,14 @@ def _get_battery(pl):
 	else:
 		pl.debug('Not using pmset: executable not found')
 
+	if sys.platform.startswith('cygwin'):
+		BATTERY_PERCENT_RE = re.compile('[0-9]+')
+		def _get_capacity(pl):
+			battery_summary = run_cmd(pl, ['WMIC', 'Path','Win32_Battery', 'GET', 'EstimatedChargeRemaining'])
+			battery_percent = BATTERY_PERCENT_RE.search(battery_summary).group(0)
+			return int(battery_percent)
+		return _get_capacity
+	
 	if sys.platform.startswith('win'):
 		# From http://stackoverflow.com/a/21083571/273566, reworked
 		try:
--- a/powerline/segments/common/bat.py
+++ b/powerline/segments/common/bat.py
@@ -95,7 +95,7 @@ def _get_battery(pl):
 	else:
 		pl.debug('Not using pmset: executable not found')
 
-	if sys.platform.startswith('win'):
+	if sys.platform.startswith('win') or sys.platform == 'cygwin':
 		# From http://stackoverflow.com/a/21083571/273566, reworked
 		try:
 			from win32com.client import GetObject
@@ -116,9 +116,14 @@ def _get_battery(pl):
 
 					return _get_capacity
 				pl.debug('Not using win32com.client as no batteries were found')
-
-		from ctypes import Structure, c_byte, c_ulong, windll, byref
-
+		if sys.platform == 'cygwin':
+			pl.debug('Using cdll to communicate with kernel32 (Cygwin)')
+			from ctypes import Structure, c_byte, c_ulong, cdll, byref
+			library_loader = cdll
+		else:
+			pl.debug('Using windll to communicate with kernel32 (Windows)')
+			from ctypes import Structure, c_byte, c_ulong, windll, byref
+			library_loader = windll
 		class PowerClass(Structure):
 			_fields_ = [
 				('ACLineStatus', c_byte),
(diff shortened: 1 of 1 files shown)


powerline/segments/common/bat.py
```diff
@@ -95,6 +95,14 @@ def _get_capacity(pl):
 	else:
 		pl.debug('Not using pmset: executable not found')
 
+	if sys.platform.startswith('cygwin'):
```
ZyX-I: Missing `executable` check. Missing `pl.debug` call telling which variant was chosen.

powerline/segments/common/bat.py
```diff
@@ -95,6 +95,14 @@ def _get_capacity(pl):
...
+		BATTERY_PERCENT_RE = re.compile('[0-9]+')
+		def _get_capacity(pl):
+			battery_summary = run_cmd(pl, ['WMIC', 'Path','Win32_Battery', 'GET', 'EstimatedChargeRemaining'])
```
ZyX-I: Invalid list formatting: comma must always be followed by one space or a line break.

powerline/segments/common/bat.py
```diff
@@ -95,6 +95,14 @@ def _get_capacity(pl):
...
+			battery_percent = BATTERY_PERCENT_RE.search(battery_summary).group(0)
+			return int(battery_percent)
+		return _get_capacity
+	
 	if sys.platform.startswith('win'):
```
ZyX-I: I am wondering whether this code can be reused first. Using `win32com.client` or ctypes must be much faster then  spawning any process.

aikeru:
Cygwin Python can't use COM. It's not impossible for a Cygwin process to call out to COM (ie cygwin ruby has win32ole) but as I did some reading this is not available in cygwin python.

I'd be happy to make the formatting changes (I'm a Python newbie) although what is meant by "executable check" and is there an example?

Calling COM from straight C is not something anyone should want to do either. This is not a trvial C program to write.

It would be faster if python in cygwin supported this. Maybe you can find something I don't know? :) I'd love to be proven wrong.

ZyX-I:
I meant that you should check whether executable is available which is done via `which`. Example is just above with `pmset`.

aikeru:
Maybe there is some WIN32 API call that Python could invoke.

aikeru:
Ah OK when I get a chance I'll update the PR :)
Thanks

powerline/segments/common/bat.py
```diff
@@ -95,6 +95,14 @@ def _get_capacity(pl):
 	else:
 		pl.debug('Not using pmset: executable not found')
 
+	if sys.platform.startswith('cygwin'):
+		BATTERY_PERCENT_RE = re.compile('[0-9]+')
+		def _get_capacity(pl):
+			battery_summary = run_cmd(pl, ['WMIC', 'Path','Win32_Battery', 'GET', 'EstimatedChargeRemaining'])
+			battery_percent = BATTERY_PERCENT_RE.search(battery_summary).group(0)
```
ZyX-I: I am wondering what this line will return if there is *no* battery. It must return `None`, **not** raise an `AttributeError` (if `re.search` have not found anything it will return `None` and `None` has no attribute `group`). I also do not know what `WMIC` will do in this case: if it exits with non-zero status (I guess unlikely) you will get exception earlier, but you should not catch it unless it *does* exit with non-zero code without battery.

aikeru: Thats a good point.. I should check this out..

aikeru:
I notice that cygwin python doesn't even seem to have win32api or win32com or anything I can find on the web that allows it to invoke WIN32 APIs without writing C/C++ code that I assume would need to be compiled OR invoking a separate process such as WMIC, .NET, WScript, etc.

I know invoking another process is less than ideal ... but is including native code that must be compiled or native assets an acceptable prerequisite for segments in powerline?
If it is, is there a convention for including native/compiled assets/source code/etc.?

Pretty disappointing that cygwin python doesn't include these things. I checked both 2.x and 3.x.

In any case, this PR gives a battery status where before there wasn't any for this platform. :)

aikeru:
OK so I consulted with someone who actually knows python. It looks like this is possible to do using ctypes. I didn't know python's ctypes were so powerful :) let me look into this further...

powerline/segments/common/bat.py
```diff
@@ -143,7 +148,7 @@ def _get_capacity(pl):
 			pl.debug('Using GetSystemPowerStatus')
 
 		return _get_capacity
-
+
```
ZyX-I: I guess this should not be here.

powerline/segments/common/bat.py
```diff
@@ -116,9 +116,14 @@ def _get_capacity(pl):
 
 					return _get_capacity
 				pl.debug('Not using win32com.client as no batteries were found')
-
-		from ctypes import Structure, c_byte, c_ulong, windll, byref
-
+		if sys.platform == 'cygwin':
+			pl.debug('Using cdll to communicate with kernel32 (Cygwin)')
+			from ctypes import Structure, c_byte, c_ulong, cdll, byref
```
ZyX-I: I would suggest moving anything, but cdll/windll import out of both branches because otherwise it looks like if you were using two identical lines. At least windll/cdll must be the first one because in this position it is easier to find differences.

aikeru:
Is there anything else that needs doing?
Maybe I can contribute to other cygwin compatibility issues (ie load avg) once this is fixed. ..

ZyX-I:
Nothing, except that there are problems with commit messages (one does not start with a capital, another is too long (1st line >70 characters)) and their amount (too many messages for code that adds a few lines).

ZyX-I:
I have left only two commits: one with WMIC (with two next commits squashed) and one with cdll (again with two squashed commits).

PR #1285, opened 2015-01-25
Add BAR support with i3 bindings

ZyX-I:
Ref #1256.

### DIFF
--- a/docs/source/usage/wm-widgets.rst
+++ b/docs/source/usage/wm-widgets.rst
@@ -44,13 +44,30 @@ Add the following to :file:`~/.config/qtile/config.py`:
        ),
    ]
 
+.. _bar-usage:
+
+LemonBoy’s bar
+=============
+
+To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
+options, for example like this::
+
+    python powerline-bar.py | bar -f "-xos4-*"
+
+to run with i3, simply ``exec`` this in i3 config file::
+
+    exec python powerline-bar.py | bar -f "-xos4-*"
+
+See the `bar documentation <https://github.com/LemonBoy/bar>`_ for more information and options.
+
 I3 bar
 ======
 
 .. note::
-    Until the patch is done in i3 a custom ``i3bar`` build called ``i3bgbar`` is 
-    needed. The source is available `in S0lll0s/i3bgbar github repository 
-    <https://github.com/S0lll0s/i3bgbar>`_.
+   As the patch to include background-colors in i3bar is likely not to be 
+   merged, it is recommended to instead run ``bar`` (see above). The source for 
+   i3bgbar is however still available `here 
+   <https://github.com/S0lll0s/i3bgbar>`_.
 
 Add the following to :file:`~/.i3/config`::
 
--- a/docs/source/usage/wm-widgets.rst
+++ b/docs/source/usage/wm-widgets.rst
@@ -47,18 +47,19 @@ Add the following to :file:`~/.config/qtile/config.py`:
 .. _bar-usage:
 
 LemonBoy’s bar
-=============
+==============
 
-To run the bar simply pipe the output of the binding script into ``bar`` and specify appropriate
-options, for example like this::
+To run the bar simply pipe the output of the binding script into ``bar`` and 
+specify appropriate options, for example like this::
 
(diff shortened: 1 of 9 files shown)


S0lll0s:
Okay, I will test everything today.

S0lll0s:
Getting `UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 30: ordinal not in range(128)` (in powerline-bar.py line 40).

That's why I had the .encode() there. py2.7.3 btw

ZyX-I:
@S0lll0s You should try with `LANG=en_US.UTF-8`. `python -c 'print(u"\xA0")'` works as long as you have correct locale.

ZyX-I:
Though it is strange you get this error at all. Without a unicode locale powerline should fall back to `ascii` theme. One thing that is missing is encoding according to the locale in `powerline-bar.py` (note: *according to the locale*, not *using utf-8*). I should ask here though whether bar expects UTF-8 (in which case `.encode('utf-8')` should happen *always* and not just on Python-2) or it expects string in the current locale (in which case there should be `.encode(get_preferred_output_encoding())`?

Also about i3bar: do you know whether it works with locale or with unicode? Because there should actually also be `.encode()`.

ZyX-I:
C code is pretty simple and expects UTF-8 as input. But it uses `uint16_t` for one character, so it obviously cannot work with unicode characters.

ZyX-I:
@S0lll0s Last commits should fix this. Also ref LemonBoy/bar#99, I am not really sure I do not have Unicode-7 characters in default theme, but I definitely do have them in `powerline_unicode7` theme (it is the very point of this theme).

S0lll0s:
Yes, everything seems to work now. I can't run the tests in the same environment though, my virtual machine's internet connection is ruined and the dependency download is too slow. If the tests work fine on your end everything should be alright though, the output you put there is correct (as indicated by my live testing)

ZyX-I:
@S0lll0s Still wondering what i3bar expects as its input. I sense I need to have UTF-8 encoding there as well.

S0lll0s:
I came up with using bar because i3bar breaks compatability constantly anyway. Some patches are "too extreme" for i3bar and i3 is pretty fast-moving so a 3rd-party patch is tiresome to sustain. i3 is using UTF-8 for everything as far as I know though, that should hold true for i3bar aswell.

ZyX-I:
@S0lll0s Python-3 is also trying hard to have everything UTF-8. But it will use current locale both for arguments and input: e.g.

```
 zyx  ~  python3.4 -c 'import sys; print (sys.argv)' a b c «»
['-c', 'a', 'b', 'c', '«»']
 zyx  ~  LANG=C python3.4 -c 'import sys; print (sys.argv)' a b c «»
['-c', 'a', 'b', 'c', '\udcc2\udcab\udcc2\udcbb']
 zyx  ~  echo «» | python3.4 -c 'import sys; print (sys.stdin.read())'
«»

 zyx  ~  echo «» | LANG=C python3.4 -c 'import sys; print (sys.stdin.read())'
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 0: ordinal not in range(128)
```

I do not see anything relevant in documentation (in fact, I only see it saying about unsupported formatting, is this `i3bgbar`-specific extension)? Though as I see it using `json.dumps` this is really not relevant because by default the result is ASCII-only. Have anybody seen non-ASCII-compatible locale being used?

S0lll0s:
The input-format of i3bgbar and i3bar is almost the same (and the differences have nothing to do with text encodings). When I wrote the i3bgbar renderer I just went with what worked fine for me (`json.dumps` did).
My locale is and was set to `en_US.UTF-8`.

And I don't really understand your last sentence?

ZyX-I:
> And I don't really understand your last sentence?

ASCII-compatible locale is a locale where all bytes in range from 0x00 to 0x7F (inclusive) are interpreted as ASCII characters. Not all encodings are like this. If you try to generate locale with such encoding `locale-gen` will complain that locale is not ISO C compliant though, but this is glibc `locale-gen`. I guess though that Python will not run on systems without ISO C compliance.

PR #1297, opened 2015-02-04
Fix socket address size on non-Linux platforms.

seanfisk:
After some debugging, I found that the Powerline client on my OS X was not connecting to the daemon. After some more debugging, it turned out that it was an incorrect `sockaddr_un` size being passed to `connect`.

This pull request should fix this for non-Linux platforms. If the platform defines `SUN_LEN`, we'll use that, otherwise fall back to the size of the struct. No changes were made for Linux.

This code was tested on OS X Mavericks (10.9) and CentOS 7.

On a side note, would you be opposed to me adding more options for debugging the clients? I know the goal is to keep them small and fast, but it was confusing for a newcomer like me to be "using the daemon" but still have a slow prompt. I'd have to think about how it could be implemented, but it would be nice to know that the client isn't connecting even when it falls back to exec'ing `powerline-render`.

Thanks!

### DIFF
--- a/client/powerline.c
+++ b/client/powerline.c
@@ -42,6 +42,26 @@ void do_write(int sd, const char *raw, size_t len) {
 	}
 }
 
+inline size_t true_sun_len(const struct sockaddr_un *ptr) {
+#ifdef __linux__
+	/* Because SUN_LEN uses strlen and abstract namespace paths begin
+	 * with a null byte, SUN_LEN is broken for these. Passing the full
+	 * struct size also fails on Linux, so compute manually. The
+	 * abstract namespace is Linux-only. */
+	if (ptr->sun_path[0] == '\0') {
+		return sizeof(ptr->sun_family) + strlen(ptr->sun_path + 1) + 1;
+	}
+#endif
+#ifdef SUN_LEN
+	/* If the vendor provided SUN_LEN, we may as well use it. */
+	return SUN_LEN(ptr);
+#else
+	/* SUN_LEN is not POSIX, so if it was not provided, use the struct
+	 * size as a fallback. */
+	return sizeof(struct sockaddr_un);
+#endif
+}
+
 #ifdef __linux__
 # define ADDRESS_TEMPLATE "powerline-ipc-%d"
 # define A +1
@@ -91,7 +111,7 @@ int main(int argc, char *argv[]) {
 	server.sun_family = AF_UNIX;
 	strncpy(server.sun_path A, address, strlen(address));
 
-	if (connect(sd, (struct sockaddr *) &server, (socklen_t) (sizeof(server.sun_family) + strlen(address) A)) < 0) {
+	if (connect(sd, (struct sockaddr *) &server, true_sun_len(&server)) < 0) {
 		close(sd);
 		/* We failed to connect to the daemon, execute powerline instead */
 		argc = (argc < NEW_ARGV_SIZE - 1) ? argc : NEW_ARGV_SIZE - 1;


client/powerline.c
```diff
@@ -90,8 +91,18 @@ int main(int argc, char *argv[]) {
...
+#elif defined SUN_LEN
+	/* If the vendor provided SUN_LEN, we may as well use it. */
+	socklen = SUN_LEN(&server);
+#else
+	/* SUN_LEN is not POSIX, so if it was not provided, use the struct size as a fallback. */
+	socklen = sizeof(struct sockaddr_un);
+#endif
 
-	if (connect(sd, (struct sockaddr *) &server, (socklen_t) (sizeof(server.sun_family) + strlen(address) A)) < 0) {
+	if (connect(sd, (struct sockaddr *) &server, socklen) < 0) {
```
ZyX-I: If you do this you need to get rid of `A` macros. It has always the same value in the place where it is currently used.

client/powerline.c
```diff
@@ -90,8 +91,18 @@ int main(int argc, char *argv[]) {
 	memset(&server, 0, sizeof(struct sockaddr_un));
 	server.sun_family = AF_UNIX;
 	strncpy(server.sun_path A, address, strlen(address));
+#ifdef __linux__
+	/* We use a socket in the abstract namespace on Linux, and because SUN_LEN uses strlen, it does not work for these. The struct size also fails, so compute manually. */
+	socklen = sizeof(server.sun_family) + strlen(address) A;
+#elif defined SUN_LEN
```
ZyX-I: My headers (Gentoo linux) do contain SUN_LEN. But this is not going to be ever used because of `#ifdef __linux__` being the first.

ZyX-I: But one needs to `#define __USE_MISC` prior to `#includ`ing `sys/un.h`. Or rather `#define` `_GNU_SOURCE` which will `#define` `_BSD_SOURCE` (it is also `#define`d by defining `_DEFAULT_SOURCE`).

seanfisk: > My headers (Gentoo linux) do contain SUN_LEN. But this is not going to be ever used because of #ifdef __linux__ being the first.

Correct. I haven't looked at the Gentoo headers, but if they are similar to Cent, `SUN_LEN` is broken for abstract namespace paths. I would like to keep the behavior for Linux the same, because it's currently working fine.

My `sys/un.h` header on OS X has this guard:
```c
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
/* actual length of an initialized sockaddr_un */
#define SUN_LEN(su) \
	(sizeof(*(su)) - sizeof((su)->sun_path) + strlen((su)->sun_path))
#endif	/* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */
```
So if we are not compiling with `_POSIX_C_SOURCE`, `SUN_LEN` should be defined.

client/powerline.c
seanfisk: Actually, it's still used in the `strncpy` line above the ifdefs. But I can move the `strncpy` line under the ifdefs if you'd like.

ZyX-I: Leave this as-is then. Though neither variant looks particularly good to me.

Since address is copied to `sun_path` can’t you move all platform-dependent code into one set of `#define`s above and define your own `SUN_LEN` (to prevent using `#undef` it should have different name)?

client/powerline.c
ZyX-I: Gentoo is using

```C
# define SUN_LEN(ptr) ((size_t) (((struct sockaddr_un *) 0)->sun_path)        \
                      + strlen ((ptr)->sun_path))
```

I think it is broken as well. Wondering what was whoever designed the first part was thinking of. If I am not mistaking this is something like `offsetof` macro, but only usable for `type []` members of the structure. I do not like messing with NULL pointers like this, this always smells like UB even if it has no UB.

client/powerline.c
seanfisk: Maybe we should just fix it once and for all [as suggested here](https://lists.debian.org/debian-glibc/2004/10/msg00135.html)? I'll give it a try.

client/powerline.c
ZyX-I: All three variants give the same result:

echo -E '#include <stddef.h>'$'\n''typedef struct {char b; char a[100];} A; main() { A *v; printf("oo:     %i\nosx:    %i\nGentoo: %i", offsetof(A, a), (int) (sizeof(*v)-sizeof(v->a)), (int) (((A*)0)->a)); }' | tcc -run -

(same for `clang -xc - && ./a.out` and `gcc`). But `offsetof` is clearly cleaner.

seanfisk:
This is in my opinion the "correct" implementation of `SUN_LEN`. This should work even if we decide to use the filesystem on Linux for any reason, at the cost of an extra branch. It's not pretty, but I tested it on both of my platforms and it works. What's your opinion?

client/powerline.c
seanfisk: The macro is pretty ugly. But I still think it's a good idea to use it when available, seeing as the `sockaddr_un` struct differs across platforms.

seanfisk:
I don't think the Travis failure has to do with my changes, but I'm not sure.

ZyX-I:
> On a side note, would you be opposed to me adding more options for debugging the clients? I know the goal is to keep them small and fast, but it was confusing for a newcomer like me to be "using the daemon" but still have a slow prompt. I'd have to think about how it could be implemented, but it would be nice to know that the client isn't connecting even when it falls back to exec'ing powerline-render.

I do not know how this should look like since it should not bother user with too much messages (i.e. like one message per each prompt) and should not bother with doing anything at all if daemon is not started. The only thing that fulfills this that I can imagine is running something like `$POWERLINE_COMMAND test-connection` *from the daemon*, **before it closes stderr** (i.e. it should run main loop prior to the part of `daemonize`, but after the first part of it); and also when daemon detects that daemon was already launched.

Wondering what was the point of daemonizing *after* creating a socket on linux.

seanfisk:
> I do not know how this should look like since it should not bother user with too much messages (i.e. like one message per each prompt) and should not bother with doing anything at all if daemon is not started.

Agreed. Maybe an error could be logged if the socket exists but the client cannot connect to it?

> The only thing that fulfills this that I can imagine is running something like `$POWERLINE_COMMAND test-connection` *from the daemon*, **before it closes stderr** (i.e. it should run main loop prior to the part of daemonize, but after the first part of it); and also when daemon detects that daemon was already launched.

Interesting. So, a self-test of sorts?

Anyway, that was just an idea. By the way, thanks for being a great maintainer :smile: Powerline is fun to use, fun to customize, and with your help, it's even fun to fix bugs. There is a lot of great code here too. I had an issue, and within 24 hours, the fix was committed. This is exactly how developing open-source software should be; it's awesome!

Thanks, ~ Sean

PR #1300, opened 2015-02-07
Another attempt to stabilize vterm tests

ZyX-I:
This variant reduces the number of commands tmux needs to run in `powerline.conf`.


ZyX-I:
Apparently this is not going to fix tests.

PR #1323, opened 2015-03-01
Add warning:regular to default color scheme

bartvm:
Tried setting up the email_imap_alert with the default scheme, but it crashed saying that the highlight color group 'email_alert' couldn't be found. This seems to fix it.


ZyX-I:
Why did you run PR to the *master* branch?

I had to force-push to destroy the merge. Never run PRs to this branch.

PR #1337, opened 2015-03-25
Support for date segment to take a timezone argument

shawncplus:
I wanted the functionally to have two clocks of different timezones in my powerline, this did the trick.

### DIFF
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -2,6 +2,7 @@
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
 from datetime import datetime
+
 from pytz import (timezone, utc)
 
 
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -3,9 +3,6 @@ from __future__ import (unicode_literals, division, absolute_import, print_funct
 
 from datetime import datetime
 
-from pytz import (timezone, utc)
-
-
 def date(pl, format='%Y-%m-%d', istime=False, tz=None):
 	'''Return the current date.
 
@@ -14,14 +11,24 @@ def date(pl, format='%Y-%m-%d', istime=False, tz=None):
 	:param bool istime:
 		If true then segment uses ``time`` highlight group.
 	:param string tz:
-		Timezone parseable by pytz
+		Timezone parseable by pytz (requires ``pytz`` module).
 
 	Divider highlight group used: ``time:divider``.
 
 	Highlight groups used: ``time`` or ``date``.
 	'''
 
-	tztime = utc.localize(datetime.utcnow()).astimezone(timezone(tz)) if tz else datetime.now()
+	if (tz):
+		try:
+			from pytz import timezone, utc
+		except:
+			tztime = datetime.now()
+			pl.error('Custom timezones require pytz module')
+		else:
+			tztime = utc.localize(datetime.utcnow()).astimezone(timezone(tz))
+	else:
+		tztime = datetime.now()
+
 	return [{
 		'contents': tztime.strftime(format),
 		'highlight_groups': (['time'] if istime else []) + ['date'],
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -2,22 +2,27 @@
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
 from datetime import datetime
+from pytz import (timezone, utc)
 
 
-def date(pl, format='%Y-%m-%d', istime=False):
+def date(pl, format='%Y-%m-%d', istime=False, tz=None):
 	'''Return the current date.
 
 	:param str format:
 		strftime-style date format string
 	:param bool istime:
 		If true then segment uses ``time`` highlight group.
+	:param string tz:
+		Timezone parseable by pytz
 
 	Divider highlight group used: ``time:divider``.
 
 	Highlight groups used: ``time`` or ``date``.
 	'''
+
+	tztime = utc.localize(datetime.utcnow()).astimezone(timezone(tz)) if tz else datetime.now()
 	return [{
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -124,7 +124,9 @@ def get_version():
 	packages=find_packages(exclude=('tests', 'tests.*')),
 	include_package_data=True,
 	zip_safe=False,
-	install_requires=[],
+	install_requires=[
+		'pytz'
+	],
```
ZyX-I: You should make this optional.

powerline/segments/common/time.py
```diff
@@ -3,21 +3,27 @@
 
 from datetime import datetime
 
+from pytz import (timezone, utc)
```
ZyX-I: It is not correct to use parenthesis here.

powerline/segments/common/time.py
```diff
@@ -3,21 +3,27 @@
...
+
+def date(pl, format='%Y-%m-%d', istime=False, tz=None):
 	'''Return the current date.
 
 	:param str format:
 		strftime-style date format string
 	:param bool istime:
 		If true then segment uses ``time`` highlight group.
+	:param string tz:
+		Timezone parseable by pytz
```
ZyX-I: Sentence does not end with a dot.

setup.py
shawncplus: Are you saying making the feature altogether optional if they have pytz installed? That makes sense.

powerline/segments/common/time.py
shawncplus: Not a python guy, mind mentioning the correct usage?

ZyX-I: It is not technically incorrect. Just you can drop the parenthesis (leaving everything else as-is) and according to the current style you should drop it unless you are going to have multiline import.

shawncplus:
I have no idea if I did the optional bits correctly but it seems work correctly with pytz installed and not and raises proper errors when trying to use unavailable functionality

powerline/segments/common/time.py
```diff
@@ -3,21 +3,34 @@
...
 		If true then segment uses ``time`` highlight group.
+	:param string tz:
+		Timezone parseable by pytz (requires ``pytz`` module).
 
 	Divider highlight group used: ``time:divider``.
 
 	Highlight groups used: ``time`` or ``date``.
 	'''
+
+	if (tz):
```
ZyX-I: You should not use parenthesis here.

powerline/segments/common/time.py
```diff
@@ -3,21 +3,34 @@
...
+		try:
+			from pytz import timezone, utc
+		except:
```
ZyX-I: Never use non-specific `except` unless your target is to catch everything and log the error.

tests/test_segments.py
```diff
@@ -684,9 +684,19 @@ class TestTime(TestCommon):
...
-			self.assertEqual(self.module.date(pl=pl, format='%H:%M', istime=True), [{'contents': '%H:%M', 'highlight_groups': ['time', 'date'], 'divider_highlight_group': 'time:divider'}])
+		try:
+			import pytz
+		except ImportError:
+			with replace_attr(self.module, 'datetime', Args(now=lambda: Args(strftime=lambda fmt: fmt))):
+				self.assertEqual(self.module.date(pl=pl), [{'contents': '%Y-%m-%d', 'highlight_groups': ['date'], 'divider_highlight_group': None}])
+				self.assertEqual(self.module.date(pl=pl, format='%H:%M', istime=True), [{'contents': '%H:%M', 'highlight_groups': ['time', 'date'], 'divider_highlight_group': 'time:divider'}])
+		else:
+			localized = Args(tzinfo=None, replace=lambda tzinfo: Args(astimezone=lambda tz: Args(strftime=lambda fmt: fmt)))
+			with replace_attr(self.module, 'utc', Args(localize=lambda time: localized)):
```
ZyX-I: I do not think tests will work.

tests/test_segments.py
```diff
@@ -684,9 +684,19 @@ class TestTime(TestCommon):
...
+				with replace_attr(self.module, 'datetime', Args(now=lambda: Args(strftime=lambda fmt: fmt), utcnow=lambda: localized)):
+					self.assertEqual(self.module.date(pl=pl), [{'contents': '%Y-%m-%d', 'highlight_groups': ['date'], 'divider_highlight_group': None}])
+					self.assertEqual(self.module.date(pl=pl, format='%H:%M', istime=True), [{'contents': '%H:%M', 'highlight_groups': ['time', 'date'], 'divider_highlight_group': 'time:divider'}])
```
ZyX-I: These two tests repeat tests above.

powerline/segments/common/time.py
```diff
@@ -3,21 +3,34 @@
...
+		Timezone parseable by pytz (requires ``pytz`` module).
 
 	Divider highlight group used: ``time:divider``.
 
 	Highlight groups used: ``time`` or ``date``.
 	'''
+
+	if (tz):
+		try:
+			from pytz import timezone, utc
```
ZyX-I: I would highly suggest importing at the top level and setting `None` in case import fails (it fails with `ImportError` and only this error should be caught) because otherwise this function is not testable.

PH111P:
While I like the idea of the `time` segment taking an optional timezone parameter, I'm not quite convinced by the implementation presented here:
* Another package is required for functionality that should be easily doable with vanilla python. In particular, specifying the timezone as ``+HHMM`` or ``-HHMM`` would allow the use of `strptime` (`%z`). 
* The `fuzzy_time` segment should also get timezone support (this should be a minor issue though)
* I find calling the additional parameter `tz` not optimal; better would be something more intuitive like `timezone`

Issue #230, opened 2013-02-17
Powerline loses colors


jakubgs:
Hello there,

After some time without updating I've run BundleUpdate today and to my surprise Powerline stopped having colors like so: ![black-n-white-powerline](https://f.cloud.github.com/assets/2212681/164924/543ba61e-78fe-11e2-95d9-38b54d606734.png)

I could get the colors by sourcing my vimrc but once I edited a new file or saved one the colors would disappear. It's all white on black.

I have no idea what might be causing this.

I use jellybeans colorscheme if that might be relevant.

Here is my vimrc: http://pastebin.com/kr9AuYgJ

SephVelut:
Issue still exists when opening new file/buffer.

`let g:miniBufExplForceSyntaxEnable = 1` fixes it.

SephVelut:
If you're using any kind of buffer autocmds (BufNewFile, BufRead, etc) then try placing `nested` between the buff discrip and the command:

`autocmd BufNewFile,BufRead *.rb nested colorscheme gotham`

PR #1342, opened 2015-04-03
Update troubleshooting.rst

SephVelut:
#230

### DIFF
--- a/docs/source/troubleshooting.rst
+++ b/docs/source/troubleshooting.rst
@@ -223,13 +223,13 @@ My vim statusline is not displayed completely and has too much spaces
 Powerline loses color after editing vimrc
 -----------------------------------------
 
-If your vimrc has something like
+If your vimrc is using buffer events with autocmd:
 
 .. code-block:: vim
 
     autocmd! BufWritePost vimrc :source ~/.vimrc
 
-to automatically source vimrc after saving it you must then add ``nested`` after 
+then to automatically source vimrc after saving it you must add ``nested`` after 
 pattern (``vimrc`` in this case):
 
 .. code-block:: vim


ZyX-I:
Commit message tells nothing about what is actually going on. Also “buffer events with autocmd” is *not* a correct description of the next line because it describes the next line too broadly and duplicates description: you were wrong about what was said after the code: it meant “If … {code} *used* to automatically source vimrc” (i.e. description of the line), not “If … {code} *then* to automatically source vimrc”.

Actually I think this troubleshooting section has only two things missing: `~/.` before `vimrc` in autocmd pattern and `used` word just after the code.

SephVelut:
Who writes commit messages for short sentence changes to a doc? The commit message is **in** the doc. 

I'm wrong? "buffer events with autocmd" *is* a correct [description](http://vimdoc.sourceforge.net/htmldoc/autocmd.html#autocmd-events). You just seem to be annoyingly evoking an overly pedantic description.

>you were wrong about what was said after the code: it meant “If … {code} used to automatically >source vimrc” (i.e. de blah blah blah blah blah etc etc

Not only a head scratcher, but insulting too. What are you talking about? 

Oh I get it, you got offended because I changed the wording you carefully constructed in commit `d5081ac`. I can't imagine any other reason for the obtuse language, prompt rejection, and bitter tone for something so simple as a sentence long rewording of a doc. Way to pitifully grasp at straws, dude.

ZyX-I:
> Who writes commit messages for short sentence changes to a doc? The commit message is in the doc.

There is no commit message in the doc. You can describe *any* change to this file as “update” and this is why I am not going to accept such commit. Even simply “Reword one sentense in troubleshooting.rst” would be better.

> I'm wrong? "events" is a correct description. You just seem to be annoyingly evoking an overly pedantic description.

It is *too broad*. It looks like if you need to use this advice for every `Buf*` event in the vimrc which is why it is *wrong*: new wording catches too much cases which *don’t need to be caught*.

> Oh I get it, you got offended because I changed the wording you carefully constructed in commit d5081ac.

You changed it so it become incorrect. I stated why I think it become incorrect.

I see there were issues, but your commit contains *too much* issues on its own for something that little and thus it was closed immediately and I did different rewording in different commit.

> I can't imagine any other reason for the obtuse language, prompt rejection, and bitter tone for something so simple as a sentence long rewording of a doc. Way to pitifully grasp at straws, dude.

Of course returning this back double and without any arguments (no, “‘events’ is a correct description” is not an argument because I clearly stated *why* it is not correct and you never answered to this “why”) is a wonderful reaction.

SephVelut:
>There is no commit message in the doc. You can describe any change to this file as “update” and this is why I am not going to accept such commit. Even simply “Reword one sentense in troubleshooting.rst” would be better.

Yeah okay, you're just grasping at straws.

>It is too broad. It looks like if you need to use this advice for every Buf* event in the vimrc which is why it is wrong: new wording catches too much cases which don’t need to be caught.

Because: 

>If your vimrc has something like

autocmd! BufWritePost vimrc :source ~/.vimrc

isn't broad at all! Worst, its absolutely vague. The original wording basically says "hey if you write bunch of code that smells like **this** then maybe you should hack around with these flags".

>Of course returning this back double and without any arguments (no, “‘events’ is a correct description” is not an argument because I clearly stated why it is not correct and you never answered to this “why”) is a wonderful reaction.

You can keep insisting that its a 'wrong' description until the cows home, mate. It won't change the fact that autocmd [events](http://vimpluginloader.sourceforge.net/doc/autocmd.txt.html#E215) for reading and writing to a file is what was being referenced. I had to add `nested` to a read autocmd for colorscheme to fix the same issue on my end `autocmd BufNewFile,BufRead *.rb nested colorscheme *****`.

ZyX-I:
> isn't broad at all! Worst, its absolutely vague.

It is not the full sentense. Last part is “used to …” which contains actual description (which is after the code because I think that this code is easier to match).

> It won't change the fact that autocmd events for reading and writing to a file is what was being referenced. I had to add nested to a read autocmd for colorscheme to fix the same issue on my end autocmd BufNewFile,BufRead *.rb nested colorscheme *****.

If you want a *generic* advice then your variant about buffer events is *too specific* and thus *still wrong* (and you needed to change *the header* as well as it only talks about vimrc resourcing). Because it applies to *every* autocommand that changes colorscheme, *not* to `Buf*` autocommands. (And specifically your code I would’ve rewritten as `autocmd FileType ruby nested colorscheme ***`: without any `Buf*` events.)

ZyX-I:
Note that I would not accept change to header because I do not expect people to connect their autocommand which resources vimrc with something described as “any autocommand that changes colorscheme”, or even “any autocommand that runs `:colorscheme` command”: purpose of the autocommand that resources vimrc does not fit into the description and it is very easy to forget that vimrc contains `:colorscheme` call at all. On the opposite side autocommand that changes colorscheme (which has this purpose) cannot be naturally connected with autocommand that resources vimrc.

Not sure whether autocommands that change colorscheme are common enough to include them in troubleshooting (though I do know that you are not the only person who uses per-filetype colorschemes).

SephVelut:
Repeat after me. **Telling. People. To. Smell. Code. For. Answers. Is. Too. Vague.** Too vague to solve any troubleshooting problems and too vague, apparently, to correctly reword the original instruction.

>If you want a generic advice then your variant about buffer events is too specific

For crying out loud.. I mentioned nothing about wanting autocmds tofor changing colorscheme in the doc. 

I didn't completely read both of your posts, as frankly they're incoherent. But if I had to guess at what you were saying, it would probably be something like this: 

"I'm so smart, I unwittingly write vague documentation, then laugh whenever someone tries to make sense of what I wrote. Then I create posts by jam packing them with as many irrelevant references - to docs and code -  as I can, to appear as if I know what I'm talking about. Even though the issue in question is as simple as a single sentence being too vague.

>your code I would’ve rewritten as autocmd FileType ruby nested colorscheme ***: without any Buf* events.

Translation: I am so smart. You had the audacity to reword my documentation. Now watch me refactor your bad code. 

Grow up. FileType breaks my colorschemes. BufRead works for multiple types, and I need the `modified` flag to be reset after executing BufNewFile. BufNewFile also works for skeleton. BufNewFile,BufRead > FileType :p 

Thanks for playing though.

ZyX-I:
> Grow up. FileType breaks my colorschemes. BufNewFile,BufRead > FileType :p

I see no reasons for FileType to break colorschemes. It does not break any of the colorschemes I have. If using FileType event breaks your colorschemes you must be doing something very stupid.

And `FileType` is run during first two events *and* when you set filetype manually or from colorschemes.

> solve any troubleshooting problems and too vague, apparently, to correctly reword the original instruction.

Did you read the troubleshooting *question*. *Nobody sane* will look at “buffer events” in a question “Powerline loses color after editing vimrc” unless he already knows the answer.

SephVelut:
>I see no reasons for FileType to break colorschemes. It does not break any of the colorschemes I have. If using FileType event breaks your colorschemes you must be doing something very stupid.

Its difficult to gauge how serious you're being, or how brain damaged. BufNewFile triggers FileType. BufNewFile also works with newly created files. BufRead allows me to specify multiple types and it waits until the contents of the file are read into memory. Now stop embarrassing yourself.

>Did you read the troubleshooting question. Nobody sane will look at “buffer events” in a question “Powerline loses color after editing vimrc” unless he already knows the answer or is simply reading everything in troubleshooting.

For the umpteenth time, the troubleshooting section in question is too vague to discern both a troubleshooting solution and a proper rewording. Entiende Usted? Bạn Hiểu Không? 你明白吗？ 

Bark once if you understand, twice if no.

ZyX-I:
> Its difficult to gauge how serious you're being, or how brain damaged. BufNewFile triggers FileType. BufNewFile also works with newly created files. BufRead allows me to specify multiple types and it waits until the contents of the file are read into memory. Now stop embarrassing yourself.

Of course, preferring *one* event over multiple ones is now called brain-damaging. You don’t need these events because FileType will be triggered by them. You also have not explained how your colorscheme breaks.

> For the umpteenth time, the troubleshooting section in question is too vague to discern both a troubleshooting solution and a proper rewording.

It is too specific and aims *only* at users who use autocommands that resource vimrc after editing it which is not your case. The event you have shown up does not launch when you do this.

---

Since you are making less and less sense as discussion continues and have shown that you are not able to understand English I am no longer interested in reading your replies here ever. Especially in this thread.

Issue #1341, opened 2015-04-02
Installation on Linux documentation AUR packages


radriaanse:
In the 'Installation on Linux' docs (https://powerline.readthedocs.org/en/master/installation/linux.html#installation-on-linux), there are links to the Arch Linux AUR packages.
These packages are outdated (release 1.3.1), and the current maintainer doesn't respond.

So until this is solved, maybe add this as a note to the documentation?, so that people won't get confused when trying to install this way.

PR #1343, opened 2015-04-04
Add note on outdated ArchLinux AUR Packages.

radriaanse:
From Issue #1341

### DIFF
--- a/docs/source/installation/linux.rst
+++ b/docs/source/installation/linux.rst
@@ -6,8 +6,8 @@ The following distribution-specific packages are officially supported, and they
 provide an easy way of installing and upgrading Powerline. The packages will 
 automatically do most of the configuration.
 
-* `Arch Linux (AUR), Python 2 version <https://aur.archlinux.org/packages/python2-powerline-git/>`_
-* `Arch Linux (AUR), Python 3 version <https://aur.archlinux.org/packages/python-powerline-git/>`_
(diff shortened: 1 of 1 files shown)


ZyX-I:
Hope Arch Linux maintainer will not forget to run the opposite PR should it continue to update the package.

Lokaltog:
I just tried to build the packages on Arch to verify #1341, and I think this commit can be reverted. Both the Arch packages currently referenced in the docs are git packages, and installing them will build the latest commit on the develop branch by default:

> ==> Finished making: python-powerline-git 2304.6b2b6d1-1 (lø. 04. april 18:05:58 +0200 2015)

There's no need to bump the version on the AUR for git packages (rebuilding will always pull the latest commit). I see no comments regarding this being an issue on the AUR either, so I'm not sure how the op on #1341 have attempted to contact the arch package maintainer about this.

There is however a [stable package](https://aur.archlinux.org/packages/python-powerline/) that's outdated at version 1.3.1, but this package is not referenced in the docs. We could possibly add a link to this package in the docs (it's an official package in the powerline/pkg-archlinux repo), and add the note about it being outdated for that package instead, but I think it would be better to wait with adding it to the docs until it's being updated consistently.

radriaanse:
Oh, this is my fault then, sorry. I didn't notice that the links in the docs refer to the git versions of the AUR packages. I was talking about the "stable" ones that aren't up to date. I contacted the maintainer by mail, because he wasn't answering the earlier comments from others.

PR #1344, opened 2015-04-04
Revert "Add note on outdated ArchLinux AUR Packages."

Lokaltog:
Reverts powerline/powerline#1343

### DIFF
--- a/docs/source/installation/linux.rst
+++ b/docs/source/installation/linux.rst
@@ -6,8 +6,8 @@ The following distribution-specific packages are officially supported, and they
 provide an easy way of installing and upgrading Powerline. The packages will 
 automatically do most of the configuration.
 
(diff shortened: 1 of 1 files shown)


radriaanse:
Why the revert? I'm waiting for a response from the current maintainer. If he doesn't respond soon, I will take over the packages.

ZyX-I:
I see that https://aur.archlinux.org/packages/py/python2-powerline-git/PKGBUILD is doing much less job then https://github.com/leycec/raiagent/blob/e896091620d77aab4e314c89d8f5ffbfd9cdd147/app-misc/powerline/powerline-9999-r20.ebuild, but more then `pip install` does. Can’t say that this means that package should be considered to be outdated or not.

Lokaltog:
The PKGBUILD needs to be updated to include the rest of the bindings, but I wouldn't call it outdated as it does install the latest available version, and installs bindings for *some* of the supported apps. It's also better than doing a `pip install`, as it installs it as a package so you'll be able to uninstall or upgrade it via the package manager. When installing powerline on Arch I'd definitely prefer these PKGBUILDs over `pip install`.

PR #1364, opened 2015-05-13
Fix dash bindings on OS X

mdeaudelin:
Fixes two errors that occur while using Powerline with the dash shell under Mac OS X.

**Problem:** Execution of the mktemp command fails.
**Cause:** mktemp requires a mandatory argument on OS X.
**Solution:** Provide a template using the -t option. While it is a deprecated option on Linux, it is still supported and provides compatibility with other systems.

**Problem:** The powerline executable fails to be called because its command line unexpectedly contains spaces.
**Cause:** The _POWERLINE_JOBS variable contains spaces because the wc command returns spaces before the number on OS X.
**Solution:** Trim the spaces before assigning the value to the variable.

### DIFF
--- a/powerline/bindings/shell/powerline.sh
+++ b/powerline/bindings/shell/powerline.sh
@@ -66,7 +66,7 @@ _powerline_set_append_trap() {
 
 _powerline_create_temp() {
 	if test -z "$_POWERLINE_TEMP" || ! test -e "$_POWERLINE_TEMP" ; then
-		_POWERLINE_TEMP="$(mktemp)"
+		_POWERLINE_TEMP="$(mktemp "${TMPDIR:-/tmp}/powerline.XXXXXXXX")"
 		_powerline_append_trap 'rm $_POWERLINE_TEMP' EXIT
 	fi
 }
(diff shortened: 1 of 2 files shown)


ZyX-I:
You should squash this commits (especially the second: it has too large commit summary, it must not extend 70 characters on its own (so commit summary is shown by github completely) and each line of the description should not exceed 80 lines).

By the way, why do you need `-t` option? I do not know how it is on Mac OS X, but on linux `-t` option does *not* specify template. It merely specifies that template should be treated relative to `${TMPDIR:-/tmp}`, so without using deprecated `-t` you can explicitly use `"${TMPDIR:-/tmp}/powerline.XXXXXXXX"`. There is also non-deprecated variant `--tmpdir` which fixes directory precedence (`-t`: $TMPDIR then `-p` then /tmp; `--tmpdir`: `-p` then $TMPDIR then /tmp).

ZyX-I:
Ah, and note that powerline runs with its own $PATH. If you need `tr` there you should explicitly symlink it: list of used executables is in tests/test_shells/test.sh.

mdeaudelin:
About the commits: Sorry about that, first public commit on GitHub so I'm not used to that yet. Will fix.

As for `mktemp`, the `--tmpdir` option is not available under OS X. There are only two possibilities: we can provide an absolute path as a template, or we can use the -t option to provide a prefix and let the command generate the template using the default temporary directory. Since the original source code used `mktemp` on its own, I assumed the intent was to let the command choose the temporary directory, which on OS X can only be done through the -t option. But, of course, as you suggested, `mktemp "${TMPDIR:-/tmp}/powerline.XXXXXXXX"` would indeed work just fine on all systems.

ZyX-I:
I reverted this fix. You must redo the PR to develop branch and *never* create PRs to master branch.

PR #1366, opened 2015-05-14
Switch the underlying i3-IPC layer and deprecate i3bgbar bindings

S0lll0s:


S0lll0s:
I need to remove the documentation for i3[bg]bar bindings and mention that you need to have `i3ipc` installed in the `bar-aint-recursive` documentation. I also want to add a segment that shows the current mode, but I don't have any more time today.

powerline/segments/i3wm.py
```diff
@@ -16,12 +16,12 @@ def calcgrp(w):
 	return group
 
 
-def workspaces(pl):
+def workspaces(pl, strip=0):
 	'''Return workspace list
```
ZyX-I: Missing argument documentation.

powerline/segments/i3wm.py
```diff
@@ -16,12 +16,12 @@ def calcgrp(w):
...
 
 	Highlight groups used: ``workspace``, ``w_visible``, ``w_focused``, ``w_urgent``
 	'''
 	return [{
-		'contents': w['name'],
+		'contents': w['name'][min(len(w['name']),strip):],
```
ZyX-I: It is rather strange to see stripping like `my workspace` → `y workspace` → ` workspace` → …: usually end of the name is stripped.

Also note that in place of stripping format argument is usually provided: e.g. `fmt.format(**w)` with default `'{name}'`.

powerline/segments/i3wm.py
```diff
@@ -1,9 +1,9 @@
 # vim:fileencoding=utf-8:noet
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
-import i3
-
+import i3ipc
```
ZyX-I: Too low number of blank lines.

powerline/segments/i3wm.py
S0lll0s: the stripping is there because i3 has a convention where you prepend numbers to your workspaces so that they stay in order ("1: www", "2: comm" etc.). I accidentally overwrote my old version and quickly recovered it like this (which is why I forgot to add documentation for `strip` again), but there probably is a smarter way.

I am not sure whether a format string is too fitting here? A regex that will be replaced with nothing would be more useful maybe (`"[0-9]+: "` in the config would strip numbers. colon and space wherever they exist)?

powerline/segments/i3wm.py
```diff
@@ -1,9 +1,9 @@
...
 
+conn = i3ipc.Connection()
```
ZyX-I: You must not create connection here. Only when actually using `workspaces` segment. And there are too low blank lines here.

powerline/segments/i3wm.py
ZyX-I: I am wondering what for this segment exists in first place. If you can switch to workspaces by number then they should not be kept as long as possible, even if this means stripping names. If you can switch to workspaces by their names then they have priority. I do not know i3, but usually you do switch to workspaces according to their numbers.

There are also no examples of regex-based stripping.

ZyX-I:
In any case I could not possibly merge this until 3.0 because dropping existing working script is backwards incompatible.

Also some other things:

1. First commit message should have `(i3-py …)` *in the commit description* and without parenthesis. Commit summaries (first lines) must not be too long for github to wrap them.
2. Why not change the code to try using `i3ipc` and fall back to `i3` if that fails? Without removing existing code, of course. No longer maintained ≠ suddenly stopped working.

S0lll0s:
`i3-py` is buggy and lacks functionality for modes. I'll try to do it in a backwards-compatible way.

powerline/segments/i3wm.py
S0lll0s: The workspace switching is completely separate from the displaying anyway, you can bind any workspace to any key(-combination). Usually they are in mod+1 through mod+9. The number prefixes are used because i3 provides special functionality for workspace handling if you follow that naming convention, including the workspaces being returned in the "correct" order (and therefore being displayed correctly in powerline). However the numbers are often not aesthetically pleasant and there should be functionality for them to be stripped.

powerline/segments/i3wm.py
S0lll0s: But then it would create a new connection every time. Should I instead use `ThreadedSegment`?

ZyX-I: Why? You need to just `global conn; if conn is None: conn = i3ipc.Connection()` and `conn = None` here. I did not say that `conn` must not be a global, I said that connection must not be created at import.

S0lll0s:
I have amended and the commits now no longer remove the i3bar bindings.
I have also added the segment I wanted to add. Is there a way to introduce vim-like modes or is that a kind of exclusive feature?

docs/source/usage/wm-widgets.rst
```diff
@@ -51,7 +51,7 @@ Add the following to :file:`~/.config/qtile/config.py`:
 
 .. _bar-usage:
 
-LemonBoy’s bar
+bar-aint-recursive
 ==============
```
ZyX-I: You must adjust the number of `=` signs. This is not a nitpicking: Sphinx *requires* header underlines to be wider or have equal width then the header.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -37,16 +39,20 @@ def render(event=None, data=None, sub=None):
...
-		import i3
-		sub = i3.Subscription(render, 'workspace')
-
-	while True:
-		start_time = monotonic()
-		render()
-		time.sleep(max(interval - (monotonic() - start_time), 0.1))
+		try:
+			import i3ipc
+			conn = i3ipc.Connection()
```
ZyX-I: This and the following should be under `else` clause of `try` block to highlight import error of which import you protect against.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -37,16 +39,20 @@ def render(event=None, data=None, sub=None):
...
+			conn.on('workspace::focus', lambda conn, evt: render())
+			conn.on('mode', lambda conn, evt: render())
+			conn.main()
+		except ImportError:
+			import i3
+			i3.Subscription(lambda evt, data, sub: print(render()), 'workspace')
+
+	while True: pass
```
ZyX-I: Note the CPU usage. You must use sleep.

powerline/segments/i3wm.py
```diff
@@ -16,12 +25,49 @@ def calcgrp(w):
...
+		conn.on('mode', callback)
+		self.thread = Thread(target=conn.main)
+		self.thread.daemon = True
+		self.thread.start()
+
+	def __call__(self, pl, default=None):
+		if self.mode == 'default':
+			return default
+		return self.mode
+
```
ZyX-I: Two blank lines.

powerline/segments/i3wm.py
```diff
@@ -16,12 +25,49 @@ def calcgrp(w):
...
 	return [{
-		'contents': w['name'],
+		'contents': w['name'][min(len(w['name']),strip):],
 		'highlight_groups': calcgrp(w)
-	} for w in i3.get_workspaces()]
+	} for w in conn.get_workspaces()]
+
+class ModeSegment(Segment):
+	argmethods = ('render')
```
ZyX-I: This is the same as `argmethods = ('r', 'e', 'n', 'd', 'e', 'r')`. Also I do not see where this method is defined.

ZyX-I:
@S0lll0s To use modes “like in vim” you must supply `mode` argument to `Powerline.render`. This is not vim-exclusive, but color modifications are only defined in Vim colorschemes.

Note: mode names must contain only unicode characters. No other restrictions are put, but I highly suggest not to use non-printable characters.

powerline/bindings/bar/powerline-bar.py
S0lll0s: you mean like

    try: import i3ipc
    except ImportError: # use i3
    else: # use i3ipc

?

powerline/segments/i3wm.py
ZyX-I: // In case you succedeed to see deleted comments: have mistaken this with `argspecobjs` method.

I do not see where this attribute is used.

powerline/bindings/bar/powerline-bar.py
ZyX-I: @S0lll0s Yes, except `import i3ipc` should be on its own line.

ZyX-I:
Another example of modes usage are shell bindings. But `mode` argument supplied by them is only used for `mode` segment because color changes are simply not defined for every group and color, except the `mode` group itself.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -20,14 +20,16 @@ def init(self):
...
+	if reschedule:
+		Timer(0.5, render, kwargs={"reschedule": True}).start()
+
 	global lock
 	with lock:
 		write(powerline.render())
 		write('\n')
 		sys.stdout.flush()
 
-
```
ZyX-I: You should not’ve deleted this line: class definitions and function definitions on the top level should be surrounded by two blank lines.

powerline/bindings/bar/powerline-bar.py
S0lll0s: Of course, just keeping it short typing on my mobile.

powerline/segments/i3wm.py
S0lll0s: My fault, left-over from an earlier attempt. I should invest more time into proof-reading.

Do you suggest I rebase and force-push until it's good or do I incrementally fix?

S0lll0s:
Is there currently a way to have a segment only show up in a certain mode? (in the user config, not by modifying that segment)

ZyX-I:
@S0lll0s Check out Vim default theme configuration file. It contains examples. Details in the reference: https://powerline.readthedocs.org/en/master/configuration/reference.html#config-themes-seg-exclude-modes.

S0lll0s:
I think I caught everything you found. I'll try to add modes to the binding next.

S0lll0s:
I'm getting a `TypeError: render() got an unexpected keyword argument 'mode'`, but according to the docs the renderer [should accept a mode argument](https://powerline.readthedocs.org/en/latest/develop/extensions.html#powerline.renderer.Renderer.render)

ZyX-I:
@S0lll0s It depends on the renderer class. Most of them override `render` method and overridden method does not accept `mode` argument if author thought it does not have to.

S0lll0s:
Also just figured that out, in this case that author is me. I'll just splat all arguments into the super calls.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -35,18 +31,37 @@ def render(event=None, data=None, sub=None):
...
+			Timer(0.5, render, kwargs={"reschedule": True}).start()
+
+		global lock
+		with lock:
+			write(line.render())
+			write('\n')
+			sys.stdout.flush()
+
+	def update(mode):
+		line.mode = mode
```
ZyX-I: I would highly suggest not to use Powerline class state. You have a number of closures here, so it is fine to save this argument in `d` dictionary:

```Patch
@@ -34,10 +34,11 @@
 	line = BarPowerline()
 	lock = Lock()
 	write = get_unicode_writer(encoding='utf-8')
+	mode_list = [None]
 
 	def render(reschedule=False):
 		if reschedule:
-			Timer(0.5, render, kwargs={"reschedule": True}).start()
+			Timer(0.5, render, kwargs={"reschedule": True, 'mode': mode_list[0]}).start()
 
 		global lock
 		with lock:
@@ -46,7 +47,7 @@
 			sys.stdout.flush()
 
 	def update(mode):
-		line.mode = mode
+		mode_list[0] = mode
 		render()
 
 	render(reschedule=True)
```

powerline/bindings/bar/powerline-bar.py
```diff
@@ -35,18 +31,37 @@ def render(event=None, data=None, sub=None):
 		help='Subscribe for i3 events.'
 	)
 	args = parser.parse_args()
-	powerline = BarPowerline()
-
-	interval = 0.5
+	line = BarPowerline()
```
ZyX-I: `powerline` name is a convention.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -35,18 +31,37 @@ def render(event=None, data=None, sub=None):
...
+			conn = i3ipc.Connection()
+			conn.on('workspace::focus', lambda conn, evt: render())
+			conn.on('mode', lambda conn, evt: update(evt.change))
+			conn.main()
 
 	while True:
-		start_time = monotonic()
-		render()
-		time.sleep(max(interval - (monotonic() - start_time), 0.1))
+		time.sleep(0.5)
```
ZyX-I: Why are you using `0.5` here? You don’t depend on the interval here, so you can use any great number like `1e10`, job will be done in other threads in any case and sleeping does not affect signal processing.

This value looks like if you *need* 0.5 for some reason because you are using it in the `Timer` call which calls for a refactoring “replace two magic numbers with a variable”.

powerline/segments/i3wm.py
```diff
@@ -16,12 +26,34 @@ def calcgrp(w):
...
+	global conn
+	if not conn: conn = i3ipc.Connection()
+
 	return [{
-		'contents': w['name'],
+		'contents': w['name'][min(len(w['name']),strip):],
 		'highlight_groups': calcgrp(w)
-	} for w in i3.get_workspaces()]
+	} for w in conn.get_workspaces()]
+
```
ZyX-I: Two blank lines.

powerline/bindings/bar/powerline-bar.py
S0lll0s: I realized I was overwriting the module, but I won't need it anymore anyway.

powerline/bindings/bar/powerline-bar.py
S0lll0s: why would I use a list though?

ZyX-I:
Is this finished?

S0lll0s:
For now it is.

PR #1372, opened 2015-05-17
Switch the underlying i3-IPC layer and deprecate i3bgbar bindings

ZyX-I:
Replaces #1366 which contains at least one error (Python-2.6 does not support `func(expr for i in iterable)`: need more parenthesis) and a number of style mistakes.

Provided tests should’ve catched the error if it was not fixed in c02516a4437c0d94f6585dce65ca349c6b399120.


S0lll0s:
As always thanks for bearing with me and fixing the trail of errors and style-felonies I leave behind.

PR #1373, opened 2015-05-19
Update reference.rst

wernight:
Clarifying a bit the confusion of lower priority vs lower priority number.

### DIFF
--- a/docs/source/configuration/reference.rst
+++ b/docs/source/configuration/reference.rst
@@ -194,7 +194,7 @@ Color definitions
 
 .. _config-colors-colors:
 
-``colors``p
+``colors``
     Color definitions, consisting of a dict where the key is the name of the 
     color, and the value is one of the following:
 
--- a/docs/source/configuration/reference.rst
+++ b/docs/source/configuration/reference.rst
@@ -194,7 +194,7 @@ Color definitions
 
 .. _config-colors-colors:
 
-``colors``
+``colors``p
     Color definitions, consisting of a dict where the key is the name of the 
     color, and the value is one of the following:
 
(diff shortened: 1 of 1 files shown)


docs/source/configuration/reference.rst
```diff
@@ -194,7 +194,7 @@ Color definitions
 
 .. _config-colors-colors:
 
-``colors``
+``colors``p
```
ZyX-I: What is `p` doing here?

docs/source/configuration/reference.rst
```diff
@@ -502,7 +502,7 @@ ascii                       Theme without any unicode characters at all
         rendered. A lower number means that the segment has a higher priority.
 
         Segments are removed according to their priority, with low priority 
-        segments being removed first.
+        segments (i.e., with a greater priority number) being removed first.
```
ZyX-I: AFAIK comma should not be here.

wernight: Yes there should be, see https://en.wiktionary.org/wiki/i.e.

docs/source/configuration/reference.rst
wernight: Hum....p

docs/source/configuration/reference.rst
ZyX-I: There are no examples of `i.e.,` in currently present documentation. (And, in fact, I almost never see comma there even though like some other Russian people I sometimes place comma far more often then needed because it is very common in our language.)

ZyX-I:
There must not be three commits with such low amount of changes and identical summary (not descriptive, BTW, though I sometimes happen to write something like “Update documentation”).

wernight:
I'd normally rebase. The web UI doesn't allow this. Web UI doesn't allow sqash commit of branch which would be another way to fix that problem.

ZyX-I:
@wernight You can clone your fork, checkout patch-1, do changes and push -f.

wernight:
Wasn't it already merged?

wernight:
Create new PR#1378

PR #1375, opened 2015-05-19
[docs] Updating default, system-wide config paths

igorsantos07:
Things now are in `{powerline_root}/config_files/{stuff}`.

I would like to add here that I've just installed Powerline through `pip` and there's no `~/powerline` nor `~/.config/powerline` in my Ubuntu box, but I'm unsure if that's an issue or it should be noted down in the docs.


ZyX-I:
Things were always there. This list was never supposed to point to repository config files.

PR #1377, opened 2015-05-22
Update reference.rst

wernight:
Clarifying a bit the confusion of lower priority vs lower priority number.


ZyX-I:
I’ve already merged both of your PR’s, just I did not use GH interface due to a number of issues that needed to be fixed.

Relevant commits are referenced in old PR’s.

ZyX-I:
Differences between my branch and your PR are mentioned in the merge commit, which is also the one referencing PR.

PR #1395, opened 2015-06-16
Support timezones without any external libraries

s0undt3ch:


powerline/segments/common/time.py
```diff
@@ -1,26 +1,38 @@
...
+		os.environ['TZ'] = timezone
+	try:
+		return [{
+			'contents': datetime.now().strftime(format),
+			'highlight_groups': (['time'] if istime else []) + ['date'],
+			'divider_highlight_group': 'time:divider' if istime else None,
+		}]
+	finally:
+		if timezone is not None and old_tz is not None:
+			os.environ['TZ'] = old_tz
```
ZyX-I: Powerline is multi-threaded (specifically, powerline-daemon is). I cannot accept code which modifies `os.environ`.

ZyX-I: Also this code is TypeError.

ZyX-I: And not tested.

s0undt3ch: Yes. I change the environment, but reset it afterwards.

ZyX-I: @s0undt3ch It is absolutely useless. It is possible that change to the environment will be reset *before* datetime runs because it will be reset *in another thread*. You cannot use locks here because `os.environ` is used in far too many places, mostly implicitly. I do not want powerline behaviour to be unstable.

powerline/segments/common/time.py
```diff
@@ -33,7 +45,7 @@ def fuzzy_time(pl, unicode_text=False):
 	'''Display the current time as fuzzy time, e.g. "quarter past six".
 
 	:param bool unicode_text:
-		If true then hyphenminuses (regular ASCII ``-``) and single quotes are 
+		If true then hyphenminuses (regular ASCII ``-``) and single quotes are
```
ZyX-I: Also you must not remove trailing spaces. They are there for a reason.

powerline/segments/common/time.py
s0undt3ch: OK. I'll see what datetime does in order to make use of that TZ environ var and I'll try to make use of that code here.

s0undt3ch:
That's an automated action from my vim

s0undt3ch:
But once rendered, since it's a new line a space will added back.

PR #1407, opened 2015-07-10
add subprocess segment

docteurklein:


ZyX-I:
1. Commit message describes adding one segment. Commit contains two.
2. No documentation.
3. redshift.py not in common. I do not see why it should have a separate module in segments because it contains nothing *extension-specific*. You can only have powerline.segments.redshift if redshift is an application that displays a statusline and you have some segments specific to this application.
4. Segments use subprocess.check_output directly. You must not do this, there is powerline.lib.shell (any alternative *must* handle decoding output and may handle some errors).
5. No tests.
6. Commit message text starts with a lowercase letter.
7. Not enough blank lines between the first function and the end of import statements.
8. There is a text literal that uses double quotes without any internal single quotes.
9. 4-space indent.

docteurklein:
thanks for the feedback. That's exactly what I needed, except your points mainly talk about (code|commit) style, not the idea by itself.
I should have been more clear from start: the only point in openeng this PR is not to be merged, but to discuss the idea of having a **generic enough** segment to be able to customize some config without having to write a whole new segment.

Thoughts?

docteurklein:
It's currently a litteral pain in the ass to customize my i3 status bar, just to add a simple info.
The "highlight group" fails lamentably and it took me like 3 hours to understand I was missing a custom colelrscheme.

Is there something obvious I missed?

ZyX-I:
@docteurklein You need to use `powerline-lint`. It should report problems with missing highlight groups right away, specifically either if you have described group in the documentation or if highlight group name is equal to segment function name. (Though I should have mentioned it in develop/segments.rst.)

Currently there is only one “generic” segment: `powerline.segments.common.env.environment`. I will merge generic segments, but if it follows the code style and has tests and documentation. Specifically for subprocess-based segments it may make sense to have *two* generic segments: one that calls other commands in the main thread and one that periodically polls (for the latter see powerline.lib.threaded module).

docteurklein:
Thanks for powerline-lint. It helps indeed!

So the idea of a "generic" segment allowing to pass arbitrary commands is something you likely see in powerline directly (if it includes tests, docs and follows CS ofc) ?

ZyX-I:
Yes. Note though that no function segments highlighting is configured via themes which makes such segment more limited. I guess it is better to make `highlight_groups` configuration option have effect on function segments (though it was disabled for a reason: many function segments will ignore this option from theme and this also has a reason: some segments consist of multiple parts with different highlighting).

But in any case I will not accept a segment that allows to specify highlight group in its arguments, *except* for the cases when it allows to pick from a limited set of groups. (Reasoning: powerline-lint should be able to check whether segment can be highlighted because without powerline-lint determining the problem is more tricky.)

Issue #1388, opened 2015-05-31
Manpage generation with python3


swiftgeek:
I had to add
"from functools import reduce"
to the beginning of /docs/source/powerline_automan.py

PR #1414, opened 2015-07-19
Fix manpage generation with python3

Siosm:
Fix manpage generation with Python 3 and allow packagers to select which version of sphinx-build (python2 ou 3) they want to use. Fixes issue #1388.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,8 +1,8 @@
 # Makefile for Sphinx documentation
-SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
-PAPER         =
-BUILDDIR      = _build
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+PAPER         ?=
+BUILDDIR      ?= _build
 
 # Internal variables
 PAPEROPT_a4     = -D latex_paper_size=a4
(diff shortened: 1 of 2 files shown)


ZyX-I:
Packagers can use `make SPHINXBUILD=sphinx-build2 man`. At least this is the case with GNU make. Are other makes different?

Siosm:
Oups, you're right I forgot to check that.
According to:
https://www.freebsd.org/doc/en_US.ISO8859-1/books/pmake/variables.html
https://www.freebsd.org/cgi/man.cgi?query=make%281%29
this looks OK for the make version shipped with FreeBSD (and probably the other *BSD). Do you have any other version in mind?

ZyX-I:
@Siosm I remember a thing named `nmake` (something windows-related, either cygwin, mingw or visual studio).

ZyX-I:
Apparently, it is part of the visual studio. At least, I see docs for it in MSDN, referencing VS 2013.

ZyX-I:
And I do not see it supporting `?=`. Reverted that commit.

Siosm:
Would you be OK if I made a commit making two different makefiles?
Or would you rather have something like http://stackoverflow.com/questions/9791127/using-make-for-cross-platform-compilation, but that I cannot test?

Siosm:
Ok, I've read you commit message and calling make with arguments works fine (I don't know why I though this would not work without '?='). You may discard my previous comment. Thanks

PR #1415, opened 2015-07-19
Workaround tmux bug: additional space in display -p '#D' output

### LINKED ISSUE #1412: Tmux right side only visible for first 3 windows
When I first start tmux, with no prior sessions, the right side of the status bar appears as normal (with uptime, time, server name, etc.). However, after I add three additional windows (for a total of four), the Powerline data changes as follows:

* If `powerline-daemon` **is** running, then the right side of the status bar changes to the first line of the powerline-daemon usage message: `usage: powerline-daemon [-h] [-r MODULE] [-w WIDTH] [--last-exit-code INT]`.
* If `powerline-daemon` **is not** running, then the right side of the status bar becomes blank.

If I kill or start powerline-daemon while inside a tmux session, the status bar will adjust to the corresponding appearance after a moment.

This is not an issue of *how many* tmux windows I have. If, after creating new windows, I switch back to one of the first three windows, the status bar goes back to normal. Switching back to window 4+, the status bar goes wrong again.

This is not an issue of the window number. If I destroy window 1 (where the status bar was correct), and then create another window, that new window will be numbered 1, but the status bar won't be correct for it. I suspect this is because the "pane ID" always increases; although window numbers get re-used, pane IDs do not.

The tmux window count applies across sessions, so if I "use up" the correctly working windows in one session, disconnect from that session, and create a new session, *all* the windows in the new session will have the wrong status bar.

Killing the tmux server (as by exiting all sessions) and restarting it makes new windows have the right status bar, but only the first three windows again.

If I run the tmux command `run-shell 'powerline tmux right'`, the output appears correct. Likewise if I run <code>env "$POWERLINE_COMMAND" $POWERLINE_COMMAND_ARGS tmux right -R pane_id=\`tmux display -p "#D"\`</code>, which is what's configured in *bindings/tmux/powerline-base.conf*. (In my environment, `POWERLINE_COMMAND` is simply `powerline`, and `POWERLINE_COMMAND_ARGS`.)

I'm using Powerline 2.2 installed with `pip3 install powerline-status` under Python 3.4 and Tmux 2.0.

What else can I do to investigate this problem?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24395018-tmux-right-side-only-visible-for-first-3-windows?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ZyX-I:
Fixes #1412

### DIFF
--- a/powerline/renderers/tmux.py
+++ b/powerline/renderers/tmux.py
@@ -63,7 +63,7 @@ class TmuxRenderer(Renderer):
 		if segment_info:
 			r.update(segment_info)
 		if 'pane_id' in r:
-			varname = 'TMUX_PWD_' + r['pane_id'].lstrip('%')
+			varname = 'TMUX_PWD_' + r['pane_id'].lstrip('% ')
 			if varname in r['environ']:
 				r['getcwd'] = lambda: r['environ'][varname]
 		r['mode'] = mode
--- a/docs/source/troubleshooting.rst
+++ b/docs/source/troubleshooting.rst
@@ -244,8 +244,8 @@ highlighting groups are usually cleared, including those defined by powerline.
 To workaround this issue powerline hooks ``Colorscheme`` event, but when you 
 source vimrc with ``BufWritePost`` (or any other) event, but without ``nested`` 
 this event is not launched. See also `autocmd-nested 
-<http://vimpluginloader.sourceforge.net/doc/autocmd.txt.html#autocmd-nested>`_ 
-Vim documentation.
+<http://vimcommunity.bitbucket.org/doc/autocmd.txt.html#autocmd-nested>`_ Vim 
+documentation.
 
 Powerline loses color after saving any file
 -------------------------------------------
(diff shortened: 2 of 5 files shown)


powerline/renderers/tmux.py
```diff
@@ -63,7 +63,7 @@ def get_segment_info(self, segment_info, mode):
 		if segment_info:
 			r.update(segment_info)
 		if 'pane_id' in r:
-			varname = 'TMUX_PWD_' + r['pane_id'].lstrip('%')
+			varname = 'TMUX_PWD_' + r['pane_id'].lstrip('%').lstrip(' ')
```
rkennedy: In the bug this is meant to work around, the spaces come first, *before* the percent sign. This code strips spaces second. Best would probably be to remove them both simultaneously: `lstrip('% ')`.

ZyX-I: You are absolutely right.

PR #1418, opened 2015-07-24
Add ac-charging indication for battery segment

xfumihiro:


powerline/segments/common/bat.py
```diff
@@ -51,10 +50,16 @@ def _get_battery(pl):
...
+					return float(
 						dbus.Interface(dev, dbus_interface=devinterface).Get(
 							devtype_name,
 							'Percentage'
+						),
+						bool(
+							dbus.Interface(dev, dbus_interface=devinterface).Get(
+								devtype_name,
+								'State'
+							) == 1
```
ZyX-I: `==` operator will already return a boolean. No need to wrap it with `bool`.

powerline/segments/common/bat.py
```diff
@@ -51,10 +50,16 @@ def _get_battery(pl):
 						pl.debug('Not using DBUS+UPower with {0}: not a power supply', devpath)
 						continue
 					pl.debug('Using DBUS+UPower with {0}', devpath)
-					return lambda pl: float(
```
ZyX-I: This function *must* return another function. I am not merging in a patch that removes caching.

powerline/segments/common/bat.py
```diff
@@ -65,13 +70,20 @@ def _get_battery(pl):
...
-					with open(cap_path, 'r') as f:
-						return int(float(f.readline().split()[0]))
-
-				return _get_capacity
-		pl.debug('Not using /sys/class/power_supply as no batteries were found')
+				with open(cap_path, 'r') as f:
+					_capacity = int(float(f.readline().split()[0]))
+
+		linux_ac_fmt = '/sys/class/power_supply/{0}/online'
+		for linux_ac in os.listdir('/sys/class/power_supply'):
```
ZyX-I: There is already one iteration over `/sys/class/power_supply` contents. Do not use two iterations.

ZyX-I: There is already an iteration over `/sys/class/power_supply`. You may use it to determine both BAT and AC directories.

powerline/segments/common/bat.py
```diff
@@ -65,13 +70,20 @@ def _get_battery(pl):
...
+			online_path = linux_ac_fmt.format(linux_ac)
+			if linux_ac.startswith('AC') and os.path.exists(online_path):
+				pl.debug('Using /sys/class/power_supply with ac {0}', linux_ac)
+				with open(online_path, 'r') as f:
+					_ac_powered = bool(f.readline())
```
ZyX-I: Also how the contents of the file looks like? If this is something like `0` or `1` then this code is incorrect: `bool("0") == bool("1") == (len("0") != 0) == (len("1") != 0) == True`.

powerline/segments/common/bat.py
```diff
@@ -189,6 +174,8 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
 		another gradient level and highlighting group, so it is OK for it to be 
 		the same as full_heart as long as necessary highlighting groups are 
 		defined.
+	:param str charging:
+		Indication of "AC charging"
```
ZyX-I: This is missing a dot. Also it should be more explicit about what kind of indication it is: “If computer is connected to a power supply this symbol is prepended to the segment.”.

ZyX-I: No, you should move `charging` to format in place of prepending it. So, two symbols: `online` and `offline` and another keyword argument to `format` string (`ac_state`: `format='{ac_state}{capacity:3.0%}'` will form a segment which never changes width: used as `format.format(ac_state=(online if ac_powered else offline), capacity=(capacity / 100.0))`; default for `offline` is simply a space).

powerline/segments/common/bat.py
```diff
@@ -220,7 +208,7 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
 		})
```
ZyX-I: If `gamify` is true then `online` or `offline` symbol should be prepended with `['battery_online', 'battery_ac_state', 'battery_gradient', 'battery']` or `['battery_offline', 'battery_ac_state', 'battery_gradient', 'battery']` highlighting.

powerline/segments/common/bat.py
```diff
@@ -196,10 +183,11 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
 	Highlight groups used: ``battery_full`` or ``battery_gradient`` (gradient) or ``battery``, ``battery_empty`` or ``battery_gradient`` (gradient) or ``battery``.
```
ZyX-I: With my suggestions regarding gamify this is how this line should look like: (just copy it literally)

    Highlight groups used: ``battery_full`` or ``battery_gradient`` (gradient) or ``battery``, ``battery_empty`` or ``battery_gradient`` (gradient) or ``battery``, ``battery_online`` or ``battery_ac_state`` or ``battery_gradient`` (gradient) or ``battery``, ``battery_offline`` or ``battery_ac_state`` or ``battery_gradient`` (gradient) or ``battery``.

tests/test_segments.py
```diff
@@ -881,10 +881,10 @@ class TestBat(TestCommon):
 	def test_battery(self):
 		pl = Pl()
 
-		def _get_capacity(pl):
-			return 86
+		def _get_battery_status(pl):
+			return 86, False
```
ZyX-I: There should be more tests:

1. Copy of the current one, but with `True` as the battery status second result.
2. Explicit `online` and `offline` keywords and `True`.
3. Explicit `online` and `offline` keywords and `False`.

ZyX-I:
And ignore most of travis failures, they are due to some zsh update (some tests did not get it for an unrelated issue with pyuv on the dependency build bot).

powerline/segments/common/bat.py
```diff
@@ -136,41 +141,19 @@ class PowerClass(Structure):
...
-	def _failing_get_capacity(pl):
-		raise NotImplementedError
-
-	try:
-		_get_capacity = _get_battery(pl)
-	except NotImplementedError:
-		_get_capacity = _failing_get_capacity
-	except Exception as e:
-		pl.exception('Exception while obtaining battery capacity getter: {0}', str(e))
-		_get_capacity = _failing_get_capacity
```
ZyX-I: This is caching. `_get_battery` is called only once and it returns some function or fails with some exception. After this `_get_capacity` function is replaced either with returned function or with `_failing_get_capacity` and bold `_get_battery` function is *never called after this*.

Caching is saving the result of calculations in order to not repeat it. Calculations are determining which method should be used for obtaining battery capacity.

powerline/segments/common/bat.py
```diff
@@ -179,32 +184,52 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
...
+			'draw_inner_divider': False,
+			'highlight_groups': ['battery_online', 'battery_ac_state', 'battery_gradient', 'battery'],
+			'gradient_level': 0,
+		})
+		ret.append({
+			'contents': '' if ac_powered else offline,
+			'draw_inner_divider': False,
+			'highlight_groups': ['battery_offline', 'battery_ac_state', 'battery_gradient', 'battery'],
+			'gradient_level': 0,
+		})
```
ZyX-I: It is bad idea to have segment with empty contents. You just need to have `'contents': online if ac_powered else offline` and below `'highlight_groups' : ['battery_online' if ac_powered else 'battery_online', …`.

powerline/segments/common/bat.py
```diff
@@ -179,32 +184,52 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
...
+		If computer is connected to a power supply this symbol is prepended to the segment.
+	:param str offline:
+		If computer is NOT connected to a power supply this symbol is prepended to the segment.
 
-	``battery_gradient`` and ``battery`` groups are used in any case, first is 
+	``battery_gradient`` and ``battery`` groups are used in any case, first is
 	preferred.
 
-	Highlight groups used: ``battery_full`` or ``battery_gradient`` (gradient) or ``battery``, ``battery_empty`` or ``battery_gradient`` (gradient) or ``battery``.
+	Highlight groups used: ``battery_full`` or ``battery_gradient`` (gradient) or ``battery``,
```
ZyX-I: No wrapping. This line is mainly for script, not for humans.

powerline/segments/common/bat.py
```diff
@@ -179,32 +184,52 @@ def battery(pl, format='{capacity:3.0%}', steps=5, gamify=False, full_heart='O',
 		Number of discrete steps to show between 0% and 100% capacity if gamify
 		is True.
 	:param bool gamify:
-		Measure in hearts (♥) instead of percentages. For full hearts 
-		``battery_full`` highlighting group is preferred, for empty hearts there
```
ZyX-I: And do not remove trailing spaces.

powerline/segments/common/bat.py
```diff
@@ -109,12 +118,8 @@ def _get_capacity(pl):
...
 				for battery in wmi.InstancesOf('Win32_Battery'):
 					pl.debug('Using win32com.client with Win32_Battery')
-
-					def _get_capacity(pl):
-						# http://msdn.microsoft.com/en-us/library/aa394074(v=vs.85).aspx
-						return battery.EstimatedChargeRemaining
-
-					return _get_capacity
+					# http://msdn.microsoft.com/en-us/library/aa394074(v=vs.85).aspx
+					return battery.EstimatedChargeRemaining, battery.BatteryStatus == 6
```
ZyX-I: This is non-function return.

Issue #1420, opened 2015-07-29
Strip domain component from username?


gitkodak:
Before I dig too far into the source, I want to know if there's already a provision to strip the domain component from the username portion of the shell.

So I'm clear: right now because I'm on a Linux VM joined to an AD domain, my username shows up as username@DOMAIN.COM in my shell prompt.  That obviously takes up a lot of space and I'd like to get that back.

Thanks!

gitkodak:
So, I attempted to add this feature.  It was a change of just a few lines, modifying the tests (which I'm sure I didn't do right) took longer than just adding the feature.

I'm pretty new at all this, it's my very first attempt at contributing to an open source project.  I have absolutely 0 training.

Anyway, the CI jobs failed.  I'm not sure what to do.  Would anyone be willing to help a newbie out?

gitkodak:
I must have really hosed things up.  On my work laptop (that is actually on a domain) the env test works, but at home it's comparing my actual environment to the test environment, and that can't be right.

After running:

python -m unittest discover tests

I get this (among other failures):


======================================================================
FAIL: test_user (test_segments.TestEnv)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kodak/development/wtf/powerline/tests/test_segments.py", line 491, in test_user
    {'contents': 'def@DOMAIN.COM', 'highlight_groups': ['user']}
AssertionError: Lists differ: [{u'contents': 'kodak', u'high... != [{u'contents': u'def@DOMAIN.CO...

First differing element 0:
{u'contents': 'kodak', u'highlight_groups': [u'user']}
{u'contents': u'def@DOMAIN.COM', u'highlight_groups': [u'user']}

- [{u'contents': 'kodak', u'highlight_groups': [u'user']}]
?                 -- ^^

+ [{u'contents': u'def@DOMAIN.COM', u'highlight_groups': [u'user']}]
?                +  ^^^^^^^^^^^^^


======================================================================

PR #1426, opened 2015-08-07
Adds feature hide_domain (requested in #1420)

gitkodak:
Set "hide_domain" to true in powerline.segments.common.env.user in
your shell theme json to hide the domain component of a username
(ex, "user@DOMAIN.COM" becomes just "user".)

Tests have been added, too, but an adult may want to double check
them.

### DIFF
--- a/powerline/segments/common/env.py
+++ b/powerline/segments/common/env.py
@@ -148,11 +148,13 @@ username = False
 _geteuid = getattr(os, 'geteuid', lambda: 1)
 
 
-def user(pl, hide_user=None):
+def user(pl, hide_user=None, hide_domain=False):
 	'''Return the current user.
 
 	:param str hide_user:
 		Omit showing segment for users with names equal to this string.
+	:param bool hide_domain:
+		Drop domain component if it exists in a username (delimited by '@').
 
 	Highlights the user with the ``superuser`` if the effective user ID is 0.
 
@@ -166,6 +168,11 @@ def user(pl, hide_user=None):
 		return None
 	if username == hide_user:
 		return None
+	if hide_domain == True:
+		try:
+			username = username[:username.index('@')]
+		except ValueError:
+			pass
 	euid = _geteuid()
 	return [{
 		'contents': username,
--- a/tests/test_segments.py
+++ b/tests/test_segments.py
@@ -472,15 +472,15 @@ class TestEnv(TestCommon):
 				pass
 
 			def username(self):
-				return 'def'
+				return 'def@DOMAIN.COM'
 
 			if hasattr(self.module, 'psutil') and not callable(self.module.psutil.Process.username):
 				username = property(username)
 
 		struct_passwd = namedtuple('struct_passwd', ('pw_name',))
 		new_psutil = new_module('psutil', Process=Process)
-		new_pwd = new_module('pwd', getpwuid=lambda uid: struct_passwd(pw_name='def'))
-		new_getpass = new_module('getpass', getuser=lambda: 'def')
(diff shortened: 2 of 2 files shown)


powerline/segments/common/env.py
```diff
@@ -148,11 +149,13 @@ def _get_user():
...
 
 
-def user(pl, hide_user=None):
+def user(pl, hide_user=None, hide_domain=False):
 	'''Return the current user.
 
 	:param str hide_user:
 		Omit showing segment for users with names equal to this string.
+	:param bool hide_domain:
+		Drop domain component if itexists in a username (delimited by '@').
```
ZyX-I: Space after `it`.

powerline/segments/common/env.py
```diff
@@ -166,6 +169,11 @@ def user(pl, hide_user=None):
 		return None
 	if username == hide_user:
 		return None
+	if hide_domain == True:
+		searchfor = None
+		searchfor = re.search('@', username)
```
ZyX-I: You *must not* use `re` for such simple cases. There are `.find` and `.index` string own methods for this case.

powerline/segments/common/env.py
```diff
@@ -166,6 +169,11 @@ def user(pl, hide_user=None):
...
+		if searchfor:
+			username, _ = username.split('@')
```
ZyX-I: After index is found there is absolutely no need to do the split: you should use the found index. Match object *also* contains index: `.end()` and `.start()` methods.

powerline/segments/common/env.py
gitkodak: When searching for documentation on using the .find() and .index() str methods, I see that these are deprecated.  Is that a concern?

powerline/segments/common/env.py
gitkodak: Something like: username[0:username.find('@')]
?

powerline/segments/common/env.py
```diff
@@ -166,6 +168,8 @@ def user(pl, hide_user=None):
 		return None
 	if username == hide_user:
 		return None
+	if hide_domain == True:
+		username = username[0:username.find('@')]
```
ZyX-I: `username[0:-1]` will remove the last character.

ZyX-I: Also I do not see them deprecated.

Note: I meant `str.find` (`str` is a type), *not* `string.find` (`string` is a module containing lots of deprecated stuff).

powerline/segments/common/env.py
```diff
@@ -166,6 +168,9 @@ def user(pl, hide_user=None):
 		return None
 	if username == hide_user:
 		return None
+	if hide_domain == True:
+		if '@' in username:
```
ZyX-I: This is double iteration over string. Use EAFP + `str.index`:

```Python
try:
    username = username[:username.index('@')]
except ValueError:
    pass
```

ZyX-I:
And squash all of the commits.

ZyX-I:
Commits 66b3248 and 0a4dbfe should not be here.

ZyX-I:
Or 4cb9b32 should not be?

gitkodak:
If it's not obvious: I have no idea what I'm doing.  This is the first time I've ever tried to contribute to a project.  I'm brand new at both python and git (I'm an operations guy trying to get more involved in the dev side of devops.)  I thought I got rid of those commits, yet here they all are.  Can you give me a hint at how to recover here?  Would it be better to abandon this branch and start over, since there's only a couple of files that have been modified?

Issue #1430, opened 2015-08-21
Rename bar (bar-aint-recursive) to lemonbar


s-ol:
The [bar](https://github.com/LemonBoy/bar) statusline has been renamed to `lemonbar` and gained popularity since (even though the repo is still called `bar` on github). I have already seen two people on reddit miss the fact that there are supported bindings.

The renderer, binding and docs section need to be renamed.
##

PR #1431, opened 2015-08-21
Rename bar (bar-aint-recursive) to lemonbar

S0lll0s:
PR for issue #1430

I sneaked in a small update to the i3wm.workspaces segment, I hope that's okay.
`powerline-lemonbar.py` got a lot more complicated, I bet you will know a cleaner way than reimplement all that in `LemonbarPowerline`?


S0lll0s:
whoops, forgot about the test cases

ZyX-I:
Yes, trailing whitespace is intentional: it indicates that `gq` should treat next line as the part of the paragraph.

Also note that I need to either defer this PR to next major release or all changes need to be backward compatible. `output` option should be part of the separate PR.

S0lll0s:
alright, I will submit the workspace segment seperately then.

S0lll0s:
@ZyX-I can you tell me what exactly is going on in the failed builds? It looks like something isn't python2 compatible, but I see no related errors in the build log.

ZyX-I:
@S0lll0s In the failing build zsh has changed its output in some of the recent development versions. I have not yet changed tests according to the new behaviour. Non-failed builds are not failing because bot-ci failed to update dependencies because pyuv package fails to build as a wheel (python binary package).

S0lll0s:
so... this is not my fault so far?

ZyX-I:
Now all non-PyPy builds should fail: I have “fixed” the bot-ci issue (actually made it ignore pyuv wheel build failure) and thus all zsh versions were updated, also changed paths to vim executables (due to migration commit names changed).

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class LemonbarPowerline(Powerline):
+	get_encoding = staticmethod(lambda: 'utf-8')
+
+	def init(self, theme=None):
+		super(LemonbarPowerline, self).init(ext='wm', renderer_module='lemonbar')
+		self.theme = theme
+
+	def create_renderer(self, load_main=False, load_colors=False, load_colorscheme=False, load_theme=False):
```
ZyX-I: Why do you have this method?

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,199 @@
...
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--width', action='store', type=int,
+		default=None, help='Maximum width the bar can occupy'
+	)
+	parser.add_argument(
+		'theme', nargs='?', default='default',
+		help='The theme to render')
```
ZyX-I: I guess I know why. Use overrides facility, *not* `create_renderer` method. Also since you need overrides implement *full* facility, from environment variables for consistency and not `--theme` argument.

S0lll0s: How would I replicate the environment-variable overrides? Or do they already work and I just need to use them?

ZyX-I: @S0lll0s Check out `powerline.commands.main.finish_args()`. It is probably better to copy arguments definition from `powerline.commands.main.get_argparser` and use this function directly. In this case you must subclass `ShellPowerline`.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,199 @@
...
+	args = parser.parse_args()
+	powerline = LemonbarPowerline(theme=args.theme)
+	lock = Lock()
+	modes = ["default"]
```
ZyX-I: Use single quotes. Also on line 172 (`"reschedule"`).

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: Also note that `--theme` and `--width` keywords must *at least* be in a separate commit.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,199 @@
...
+			try:
+				renderer = Renderer(**self.renderer_options)
+			except Exception as e:
+				self.exception('Failed to construct renderer object: {0}', str(e))
+				if not hasattr(self, 'renderer'):
+					raise
+			else:
+				self.renderer = renderer
+
+if __name__ == '__main__':
```
ZyX-I: I would highly suggest moving *everything*, but a single function `main` out to `powerline/commands/lemonbar.py`. Reasoning: this makes it easy to create a compatibility script `powerline-bar.py`. (Symlinks and use of `subprocess.Popen` or equivalents in compatibility script is not that good idea because most files in `bindings` directory are supposed to be movable.)

Since arguments definition is getting bolder it should be in `powerline.commands.lemonbar.get_argparser`, similar to `powerline.commands.main` for the same reason: make it possible to create documentation and man page if needed.

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: (Actually better not copy, but move part of `powerline.commands.main.get_argparser` to a separate function.)

S0lll0s:
Almost completely forgot about this because a bunch of other stuff came up in the last weeks, I'll read everything again and try to whip something up within the next few days hopefully.

S0lll0s:
So, you suggest that these things happen, right?

1. I refractor almost everything in `powerline-bar.py` into a `powerline.commands.lemonbar`
2. I continue to provide a compatibility script `powerline-bar.py` but also create a new `powerline-lemonbar.py`
3. with the next major release the `powerline-bar.py` binding and directory are dropped

how should this be handled in the documentation?

ZyX-I:
Yes, you are right. Documentation does not have to be backward-compatible, you can simply omit the deprecated names.

ZyX-I:
It also makes sense to make `powerline-bar.py` use `pl.warn()` to issue a deprecation warning. When I previously deprecated common segments warning was given by powerline-lint, but it cannot lint powerline-bar usage.

S0lll0s:
I'm not sure whether that first commit should be on it's own, in the same commit or in a different PR entirely - it doesn't do anything on it's own but it's "an i3 thing", not lemonbar-specific.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,42 +2,45 @@
...
+		default=None, help='Maximum width the bar can occupy'
+	)
+	parser.add_argument(
+		'theme', nargs='?', default='default',
+		help='The theme to render')
 	args = parser.parse_args()
-	powerline = BarPowerline()
+	powerline = LemonbarPowerline(theme=args.theme)
+	powerline.update_renderer()
+	powerline.logger.warn("the 'bar' bindings are deprecated, please switch to 'lemonbar'")
```
ZyX-I: Do not use `powerline.logger`, it is for internal use only. `powerline.pl` must be used instead.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,42 +2,45 @@
...
 	lock = Lock()
-	modes = ['default']
+	modes = ["default"]
```
ZyX-I: Revert.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,42 +2,45 @@
...
 	write = get_unicode_writer(encoding='utf-8')
 
 	def render(reschedule=False):
 		if reschedule:
-			Timer(0.5, render, kwargs={'reschedule': True}).start()
+			Timer(0.5, render, kwargs={"reschedule": True}).start()
```
ZyX-I: Revert.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+        elif 'DVI' in screen['name'] or 'VGA' in screen['name']:
+            theme_name = 'dvi'
+
+        line = LemonbarPowerline(theme=theme_name, output=screen['name'])
+        process = subprocess.Popen("bar-aint-recursive -g {}x16+{} -f 'Source Code Pro for Powerline-9,Icon-9'".format(screen['width'], screen['x']), shell=True, stdin=subprocess.PIPE)
+        powerlines.append((line, process, int(screen['width'])/5))
+
+
+    lock = Lock()
+    modes = ["default"]
```
ZyX-I: Single quotes.

powerline/segments/i3wm.py
```diff
@@ -28,7 +29,8 @@ def workspaces(pl, only_show=None, output=None, strip=0):
 		are shown.
 
 	:param str output:
-		If specified, only workspaces on this output are shown.
+		If specified, only workspaces on this output are shown. If unspecified, 
+		may be set in ``segment_info``.
```
ZyX-I: `segment_info` is for developers. This documentation is for users. One should say or reference how this value may appear in `segment_info`, but *not* talk about `segment_info` itself.

ZyX-I: If you check documentation on readthedocs you will see that segment documentation never contains `pl` or `segment_info` in the argument list. It is removed by Powerline mocked version of autodoc, intentionally.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+from threading import Lock, Timer
+from argparse import ArgumentParser
+
+from powerline.lemonbar import LemonbarPowerline
+from powerline.lib.monotonic import monotonic
+from powerline.lib.dict import mergedicts
+
+
+if __name__ == '__main__':
+    parser = ArgumentParser(description='Powerline BAR bindings.')
```
ZyX-I: By the way, I see spaces here.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import (unicode_literals, division, absolute_import, print_function)
```
ZyX-I: Missing part of the header: `# vim:fileencoding=utf-8:noet`.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+        if 'HDMI1' in screen['name']:
+            theme_name = 'hdmi'
+        elif 'DVI' in screen['name'] or 'VGA' in screen['name']:
+            theme_name = 'dvi'
+
+        line = LemonbarPowerline(theme=theme_name, output=screen['name'])
+        process = subprocess.Popen('bar-aint-recursive -g {}x16+{} -f "Source Code Pro for Powerline-9,Icon-9"'.format(screen['width'], screen['x']), shell=True, stdin=subprocess.PIPE)
+        powerlines.append((line, process, int(screen['width'])/5))
+
+
```
ZyX-I: Too many blank lines.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+    for screen in active_screens:
+        theme_name = 'default'
+        if 'HDMI1' in screen['name']:
+            theme_name = 'hdmi'
+        elif 'DVI' in screen['name'] or 'VGA' in screen['name']:
+            theme_name = 'dvi'
+
+        line = LemonbarPowerline(theme=theme_name, output=screen['name'])
+        process = subprocess.Popen('bar-aint-recursive -g {}x16+{} -f "Source Code Pro for Powerline-9,Icon-9"'.format(screen['width'], screen['x']), shell=True, stdin=subprocess.PIPE)
```
ZyX-I: Do not use subprocess.Popen with string argument. Only list arguments are accepted.

ZyX-I: (And this line is *way* too long.)

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
S0lll0s: ah damn yeah, I started this as just a "personal" file. Actually I need to remove the font settings aswell as they are personal preference/settings here.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+        'LVDS1': 'default',
+        'HDMI1': 'hdmi',
+        'VGA1': 'hdmi'
+    }
+
+    active_screens = [
+        dict(zip(['name', 'width', 'height', 'x', 'y'], values))
+        for values in re.findall(
+            '^(\w+) connected (\d+)x(\d+)\+(\d+)\+(\d+)',
+            subprocess.check_output('xrandr -q', shell=True).decode(),
```
ZyX-I: Same: no string in `check_output` arguments.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+    for screen in active_screens:
+        theme_name = 'default'
+        if 'HDMI1' in screen['name']:
+            theme_name = 'hdmi'
+        elif 'DVI' in screen['name'] or 'VGA' in screen['name']:
+            theme_name = 'dvi'
+
+        line = LemonbarPowerline(theme=theme_name, output=screen['name'])
+        process = subprocess.Popen('bar-aint-recursive -g {}x16+{} -f "Source Code Pro for Powerline-9,Icon-9"'.format(screen['width'], screen['x']), shell=True, stdin=subprocess.PIPE)
+        powerlines.append((line, process, int(screen['width'])/5))
```
ZyX-I: Spaces around `/`.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,84 @@
...
+    themes = {
+        'LVDS1': 'default',
+        'HDMI1': 'hdmi',
+        'VGA1': 'hdmi'
+    }
+
+    active_screens = [
+        dict(zip(['name', 'width', 'height', 'x', 'y'], values))
+        for values in re.findall(
+            '^(\w+) connected (\d+)x(\d+)\+(\d+)\+(\d+)',
```
ZyX-I: I would actually suggest using named capturing groups and [groupdict](https://docs.python.org/2.6/library/re.html?highlight=re#re.MatchObject.groupdict) in place of dict+zip.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,85 @@
...
+
+from powerline.lemonbar import LemonbarPowerline
+from powerline.lib.monotonic import monotonic
+from powerline.lib.dict import mergedicts
+
+
+if __name__ == '__main__':
+	parser = ArgumentParser(description='Powerline BAR bindings.')
+	parser.add_argument(
+                '--i3', action='store_true',
```
ZyX-I: I still see spaces here. And additionally *too much* spaces.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,31 +2,34 @@
...
 		'--i3', action='store_true',
 		help='Subscribe for i3 events.'
 	)
+	parser.add_argument(
+		'--width', action='store', type=int,
+		default=None, help='Maximum width the bar can occupy'
+	)
+	parser.add_argument(
+		'theme', nargs='?', default='default',
+		help='The theme to render')
```
ZyX-I: My argument regarding overrides applies here as well (also I do not see what is the reasoning behind adding functionality to deprecated file).

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,63 @@
...
+	parser.add_argument(
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--width', action='store', type=int,
+		default=None, help='Maximum width the bar can occupy'
+	)
+	parser.add_argument(
+		'theme', nargs='?', default='default',
```
ZyX-I: This code leads to theme from configuration file being completely ignored. I cannot accept this for this reason also.

ZyX-I: (And `nargs='?'` means that you can either write `… --theme=mytheme` or `--theme`. What is the second variant supposed to mean?)

S0lll0s:
did I catch everything?

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+	parser = ArgumentParser(
+		description='Powerline BAR bindings.',
+		usage='%(prog)s [-h] [--i3] [--height HEIGHT] [--theme THEME [THEME ...]] -- args')
+	parser.add_argument(
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--height', default='',
+		help='Bar height'
```
ZyX-I: Missing sentense end.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+	)
+	parser.add_argument(
+		'--theme', '-t',
+		default=[], nargs='+',
+		help='makes all outputs that containing a string use the specified theme. Format like "out=theme"'
+	)
+	parser.add_argument(
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
```
ZyX-I: Missing sentense end.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--height', default='',
+		help='Bar height'
+	)
+	parser.add_argument(
+		'--theme', '-t',
+		default=[], nargs='+',
+		help='makes all outputs that containing a string use the specified theme. Format like "out=theme"'
```
ZyX-I: Inconsistent documentation, missing sentence end. Also I will not accept this as well, use overrides. If you need output-specific themes then this is needed to be a local theme (e.g. `"local_themes": {"{output}": "{theme_name}"}` in `config.json/ext/wm`) like you may see in vim and shell bindings, *not* `--theme output=theme` argument.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
+	)
+	args = parser.parse_args()
+	theme_rules = {}
+	for theme in args.theme:
+		out, theme = theme.split('=', 1)
+		theme_rules[out] = theme
+
+	active_screens = [match.groupdict() for match in re.finditer(
```
ZyX-I: `match.groupdict()` should be on the next line *or* next line should be aligned with `match`. Note that alignment is using spaces and it is not preferred.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,63 @@
...
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--width', action='store', type=int,
+		default=None, help='Maximum width the bar can occupy'
+	)
+	parser.add_argument(
+		'theme', default=None,
+		help='The theme to render')
```
ZyX-I: Same thing: use overrides from environment if you are sure you need this at all.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+	modes = ['default']
+
+	def render(reschedule=False):
+		if reschedule:
+			Timer(0.5, render, kwargs={'reschedule': True}).start()
+
+		global lock
+		with lock:
+			for line, process, width in powerlines:
+				process.stdin.write(line.render(mode=modes[0], width=width).encode('utf-8')+b'\n')
```
ZyX-I: Spaces around `+`.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+
+from powerline.lemonbar import LemonbarPowerline
+from powerline.lib.monotonic import monotonic
+from powerline.lib.dict import mergedicts
+
+
+if __name__ == '__main__':
+	parser = ArgumentParser(
+		description='Powerline BAR bindings.',
+		usage='%(prog)s [-h] [--i3] [--height HEIGHT] [--theme THEME [THEME ...]] -- args')
```
ZyX-I: Why you are not letting argparse construct the usage on its own?

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,63 @@
...
+
+if __name__ == '__main__':
+	parser = ArgumentParser(description='Powerline lemonbar bindings.')
+	parser.add_argument(
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--width', action='store', type=int,
+		default=None, help='Maximum width the bar can occupy'
```
ZyX-I: Dot at the end.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
```diff
@@ -0,0 +1,94 @@
...
+		command = ['bar-aint-recursive', '-g', '{}x{}+{}'.format(screen['width'], args.height, screen['x'])] + args.args[1:]
+		process = subprocess.Popen(command, stdin=subprocess.PIPE)
+		powerlines.append((line, process, int(screen['width']) / 5))
+
+	lock = Lock()
+	modes = ['default']
+
+	def render(reschedule=False):
+		if reschedule:
+			Timer(0.5, render, kwargs={'reschedule': True}).start()
```
ZyX-I: Update interval is definitely something that deserves to be in command-line argument, unlike `--theme`.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
```
ZyX-I: By the way, what’s the point in this script at all? `powerline-lemonbar-multimonitor` seems to be capable of handling single-monitor configuration fine.

docs/source/usage/wm-widgets.rst
```diff
@@ -50,24 +50,24 @@ Add the following to :file:`~/.config/qtile/config.py`:
...
+To run the bar simply pipe the output of the binding script into ``lemonbar`` and 
 specify appropriate options, for example like this::
 
-    python /path/to/powerline/bindings/bar/powerline-bar.py | bar
+    python /path/to/powerline/bindings/lemonbar/powerline-lemonbar.py | lemonbar
 
 to run with i3, simply ``exec`` this in i3 config file::
 
-    exec python /path/to/powerline/bindings/bar/powerline-bar.py --i3 | bar
+    exec python /path/to/powerline/bindings/lemonbar/powerline-lemonbar.py --i3 | lemonbar
```
ZyX-I: Missing `powerline-lemonbar-multimonitor.py` reference.

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
ZyX-I: (I have also one question: how is this working in case e.g. when you brought a notebook to home and connected a new monitor. *After* script was run.)

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
S0lll0s: there is also no preexisting facility for per-monitor theme configuration. The reason there is a `--theme` in the non-multimonitor version is the same, if you have multiple monitors an override is the only way you can have multiple themes in the same extension at the same time, and that was definetely needed for the workspace segment prior to the `segment_info['output']`. In the `powerline-lemonbar.py` the `--theme` _could_ be removed now because the use-case it was intended for is covered by `…-multimonitor.py` but I am not sure whether there are other reasons someone may need it.

ZyX-I: @S0lll0s For per-monitor theme configuration there *is* facility: it is local themes. In case somebody needs overriding its configuration the whole thing needs to be written to accept overrides from environment variables because this is *consistent* with other bindings which do not have something more usable (namely shell and tmux, vim and ipython are configured using existing vim and ipython own configuration facilities) and also more flexible. Though you will have to inherit from `ShellPowerline` then and use `powerline.commands.main.finish_args`. It makes sense to move https://github.com/ZyX-I/powerline/blob/4382993d43057ed46bc5d6cc1f18affb5a0e6d39/powerline/commands/main.py#L116-L134 and https://github.com/ZyX-I/powerline/blob/4382993d43057ed46bc5d6cc1f18affb5a0e6d39/powerline/commands/main.py#L144-L149 to a separate function for this purpose, because shell also allows overriding from command-line (though this was officially deprecated, I am not going to remove it ever).

ZyX-I: For local themes one needs

1. Linter update: https://github.com/ZyX-I/powerline/blob/59e66dca32f2bf17cb7cabec80b3c8d729795266/powerline/lint/__init__.py#L128. Three examples above, most likely you need something like

    	wm=ext_spec().update(
			components=gen_components_spec('statusline', 'tabline').optional(),
			local_themes=Spec(	).unknown_spec(
				Spec().re({place regex that matches possible output names here}),
				ext_theme_spec()
			),
		).optional(),
2. Documentation update below [this](https://github.com/ZyX-I/powerline/blob/59e66dca32f2bf17cb7cabec80b3c8d729795266/docs/source/configuration/reference.rst#L183) paragraph.

Cannot remember anything else.

powerline/bindings/lemonbar/powerline-lemonbar.py
S0lll0s: actually that's right, it never occurred to me because I had the old one laying around for a long time and wrote the multimonitor one from scratch just now

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
S0lll0s: I'll look into this

powerline/bindings/lemonbar/powerline-lemonbar-multimonitor.py
ZyX-I: I mean, this script seems to be ignoring cases when monitor is added or removed.

S0lll0s: Yeah, it doesn't react to changes in configuration. That would mean more dependencies and probably be even less universal than `xrandr` already is.

Also usually you have to reconfigure the monitors with some sort of script anyway (like xrandr) when you plug/unplug them and often you need to do more things like reapplying the wallpaper etc. When that happens you can simply restart the script and it will autodetect again.

S0lll0s:
removed the `--theme` argument and overrides and got `local_themes` to work, linting seems to be working.

S0lll0s:
I'm not sure what exactly is going on in the tests though, I think `subprocess.check_output` isn't available in all python versions but the ascii/utf8 error in `test_lemonbar_escape` I do not understand.

ZyX-I:
@S0lll0s Given where error occures it is a bug in unittest2: there is some error, but this error contains UTF-8 characters and thus unittest2 cannot print it. So you have a error in test, not necessary containing any UTF-8 problems, which is simply not displayed properly.

ZyX-I:
@S0lll0s Check Python-3.5 test output, it contains actual error. Though I think you should be able to reproduce this locally.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+	parser = ArgumentParser(
+		description='Powerline BAR bindings.',
+		usage='%(prog)s [-h] [--i3] [--height HEIGHT] [--interval INTERVAL] -- args')
+	parser.add_argument(
+		'--i3', action='store_true',
+		help='Subscribe for i3 events.'
+	)
+	parser.add_argument(
+		'--height', default='',
+		help='Bar height'
```
ZyX-I: This and the following two sentenses are missing dots at the end.

powerline/lint/__init__.py
```diff
@@ -126,6 +126,12 @@ def load_json_config(config_file_path, load=load, open_file=open_file):
 				select=ext_theme_spec(),
 			),
 		).optional(),
+		wm=ext_spec().update(
+			local_themes=Spec().unknown_spec(
+				Spec().re('[A-Z0-9-]+'),
```
ZyX-I: `[A-Z0-9-]+` literally means “starts with a latin capital letter, digit or hyphen-minus” and looks like “contains latin capital letter, digit or hyphen-minus”. You are missing anchors: `^` and `$`.

powerline/config_files/config.json
```diff
@@ -46,7 +46,9 @@
 		},
 		"wm": {
 			"colorscheme": "default",
-			"theme": "default"
+			"theme": "default",
+      "local_themes": {
+      }
```
ZyX-I: Spaces.

ZyX-I: Also this change without actually adding any local themes is not needed, revert.

tests/test_configuration.py
```diff
@@ -825,25 +825,25 @@ def tearDownClass(cls):
...
 					powerline,
 					'%{l}%{F#ffc00000}%{B#ff008000}%{+u} A%{F-B--u}%{F#ff008000}%{B#ffc00000}>>%{F-B--u}%{F#ff008000}%{B#ffc00000}B%{F-B--u}%{F#ffc00000}>>%{F-B--u}%{r}%{F#ffc00000}<<%{F-B--u}%{F#ff804000}%{B#ffc00000}%{+u}C%{F-B--u}%{F#ff0000c0}%{B#ffc00000}<<%{F-B--u}%{F#ff008000}%{B#ff0000c0}D %{F-B--u}'
 				)
 
 	@with_new_config
-	def test_bar_escape(self, config):
+	def test_lemonbar_escape(self, config):
 		import powerline as powerline_module
 		config['themes/wm/default']['segments']['left'] = (
 			highlighted_string('%{asd}', 'hl1'),
```
ZyX-I: Test failure in https://travis-ci.org/powerline/powerline/jobs/100270461 looks like you have lost escaping of per cent sign somewhere.

powerline/renderers/bar.py
```diff
@@ -1,45 +0,0 @@
...
-
-class BarRenderer(Renderer):
-	'''bar (bar ain't recursive) renderer
-
-
-	See documentation of `bar <https://github.com/LemonBoy/bar>`_ and :ref:`the usage instructions <bar-usage>`
-	'''
-
-	character_translations = Renderer.character_translations.copy()
-	character_translations[ord('%')] = '%%'
```
ZyX-I: These two lines are missing, and this was caught by the test.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+	)
+	parser.add_argument(
+		'--interval', '-i',
+		default=0.5,
```
ZyX-I: `-i` argument is a string with such code, *not* float like it should be. This is not going to work if you supply this argument on the command-line.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+		help='Refresh interval'
+	)
+	parser.add_argument(
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
+	)
+	args = parser.parse_args()
+
+	active_screens = [match.groupdict() for match in re.finditer(
```
ZyX-I: And still this should be on its own line or the following lines should be aligned (*not* indented).

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+		'^(?P<name>\w+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
```
ZyX-I: Previous complaint is invalid: I missed that you align here a `finditer` argument and not a simple part of the list comprehension expression. So this code is fine.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+	modes = ['default']
+
+	def render(reschedule=False):
+		if reschedule:
+			Timer(args.interval, render, kwargs={'reschedule': True}).start()
+
+		global lock
+		with lock:
+			for line, process, width in powerlines:
+				process.stdin.write(line.render(mode=modes[0], width=width).encode('utf-8')+b'\n')
```
ZyX-I: But spaces around `+` are still missing.

docs/source/usage/wm-widgets.rst
ZyX-I: Documentation is now invalid because the way lemonbar is invoked changed.

By the way, you say “lemonbar” here, but in the code it is “bar-aint-recursive”. Why?

powerline/segments/i3wm.py
```diff
@@ -45,6 +47,8 @@ def workspaces(pl, only_show=None, output=None, strip=0):
 		else:
 			conn = i3ipc.Connection()
 
+	output = output or ('output' in segment_info and segment_info['output'])
```
ZyX-I: Do not use weird construct `key in dict and dict[key]`, there is `dict.get(key, False)` which replicates the behaviour exactly (though you probably meant `dict.get(key)` which is the same as `dict.get(key, None)`).

docs/source/configuration/reference.rst
```diff
@@ -187,6 +187,9 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
     ``out`` and ``rewrite`` prompts (refer to IPython documentation for more 
     details) while ``in`` prompt is the default.
 
+    For the wm ext, it is a dictionary ``{output : theme_name}`` that maps
```
ZyX-I: Remove `ext,`.

docs/source/configuration/reference.rst
```diff
@@ -187,6 +187,9 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
     ``out`` and ``rewrite`` prompts (refer to IPython documentation for more 
     details) while ``in`` prompt is the default.
 
+    For the wm ext, it is a dictionary ``{output : theme_name}`` that maps 
+    the ``xrandr`` output names to the local theme to substitute on that output.
```
ZyX-I: “themes to use” (plural (to match “names”) and “use”), not “theme to substitute” (singular and “substitute”).

powerline/lemonbar.py
```diff
@@ -0,0 +1,23 @@
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+from powerline import Powerline
+from powerline.lib.dict import mergedicts
+
+
+class LemonbarPowerline(Powerline):
```
ZyX-I: Missing `get_encoding` method. Not sure whether or not it is intentional, but without this method locale encoding will be used.

powerline/segments/i3wm.py
```diff
@@ -28,7 +29,8 @@ def workspaces(pl, only_show=None, output=None, strip=0):
 		are shown.
 
 	:param str output:
-		If specified, only workspaces on this output are shown.
+		If specified, only workspaces on this output are shown. If unspecified, 
+		may be set by the lemonbar renderer and bindings.
```
ZyX-I: Not very useful: should say what it may be set to.

powerline/lemonbar.py
S0lll0s: isn't the default implementation okay? I wouldn't know where I should get it from except from the environment locale and thats what `get_preffered_output_encoding` does, no?

ZyX-I: @S0lll0s I do not know. If lemonbar expects locale encoding in the input you should live this as-is. If lemonbar expects UTF-8 only you should define this method to force UTF-8.

powerline/lint/__init__.py
```diff
@@ -126,6 +126,12 @@ def load_json_config(config_file_path, load=load, open_file=open_file):
...
+				ext_theme_spec()
+			)
+		).optional()
```
ZyX-I: And missing trailing comma.

powerline/renderers/bar.py
S0lll0s: how does this cause an encoding error though? I only added the `ord('%')` to escape it for lemonbar, but that wasn't actually needed iirc. Or I am confusing something here.

ZyX-I: @S0lll0s Why do you think it is not needed? Literal `%{r}` is a *valid* segment contents and if you do not escape `%` it will be confused with `%{r}` that starts right side in place of displaying it literally.

ZyX-I: And in the message there are unicode characters: all spaces are being replaced with non-breaking spaces. This is not in the test code or in assertion error, but this is in stack trace which shows you the source code: https://github.com/ZyX-I/powerline/blob/59e66dca32f2bf17cb7cabec80b3c8d729795266/tests/test_configuration.py#L191.

docs/source/configuration/reference.rst
```diff
@@ -187,6 +187,9 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
     ``out`` and ``rewrite`` prompts (refer to IPython documentation for more 
     details) while ``in`` prompt is the default.
 
+    For wm, it is a dictionary ``{output : theme_name}`` that maps the
```
ZyX-I: AFAIK it is incorrect to place comma here.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+	)
+	args = parser.parse_args()
+
+	active_screens = [match.groupdict() for match in re.finditer(
+		'^(?P<name>\w+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
+		subprocess.check_output(['xrandr', '-q']).decode(),
+		re.MULTILINE
+	)]
+
+	powerlines = []
```
ZyX-I: You need only *one* powerline, since themes are now selected using local_themes facility. Local theme should be selected not at powerline creation, but at render call. And also …

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,85 @@
...
+	)
+	parser.add_argument(
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
+	)
+	args = parser.parse_args()
+
+	active_screens = [match.groupdict() for match in re.finditer(
+		'^(?P<name>\w+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
+		subprocess.check_output(['xrandr', '-q']).decode(),
```
ZyX-I: … this is not present in 2.6, so you need powerline.lib.shell.run_cmd, for which you need PowerlineLodder instance which can be obtained from the only LemonbarPowerline instance created.

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: Basically this should be something like
```Patch
powerline = LemonbarPowerline()
```
…
```Patch
+	screens = []
+
+	for screen in active_screens:
+		command = ['bar-aint-recursive', '-g', '{}x{}+{}'.format(screen['width'], args.height, screen['x'])] + args.args[1:]
+		process = subprocess.Popen(command, stdin=subprocess.PIPE)
+		powerlines.append((screen['name'], process, int(screen['width']) / 5))
```
…
```Patch
+		with lock:
+			for output, process, width in powerlines:
+				process.stdin.write(powerline.render(mode=modes[0], width=width, matcher_info=output).encode('utf-8') + b'\n')
```

powerline/lemonbar.py
```diff
@@ -0,0 +1,23 @@
...
+	def init(self, output=None):
+		super(LemonbarPowerline, self).init(ext='wm', renderer_module='lemonbar')
+		self.output = output
+
+	def render(self, *args, **kwargs):
```
ZyX-I: This method should be missing completely.

powerline/lemonbar.py
```diff
@@ -0,0 +1,23 @@
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+from powerline import Powerline
+from powerline.lib.dict import mergedicts
+
+
+class LemonbarPowerline(Powerline):
+	def init(self, output=None):
```
ZyX-I: This method should be missing output argument, as well as `self.output = output`.

powerline/renderers/lemonbar.py
```diff
@@ -0,0 +1,59 @@
...
+		if bg is not None:
+			if bg is not False and bg[1] is not False:
+				text += '%{{B#ff{0:06x}}}'.format(bg[1])
+
+		if attrs & ATTR_UNDERLINE:
+			text += '%{+u}'
+
+		return text + contents + '%{F-B--u}'
+
+	def render(self, output=None, *args, **kwargs):
```
ZyX-I: And this should be missing `output` argument as well, taking output for `segment_info` from `kwargs.get('matcher_info')`.

powerline/renderers/lemonbar.py
```diff
@@ -0,0 +1,59 @@
...
+		self.segment_info['output'] = output
+		return '%{{l}}{0}%{{r}}{1}'.format(
+			super(LemonbarRenderer, self).render(side='left', matcher_info=output, *args, **kwargs),
```
ZyX-I: While here explicit `matcher_info` is not needed.

powerline/renderers/lemonbar.py
```diff
@@ -36,10 +37,26 @@ def hl(self, contents, fg=None, bg=None, attrs=None):
 		return text + contents + '%{F-B--u}'
 
 	def render(self, *args, **kwargs):
+		self.segment_info['output'] = kwargs.get('matcher_info')
```
ZyX-I: Missed that modifying `self.segment_info` is not thread-safe. With default implementation of `Renderer.get_segment_info()` method since you do not supply `segment_info` in kwargs this line should be missing and next lines should contain `segment_info={'output': kwargs.get('matcher_info')}`.

S0lll0s: shouldn't I merge `self.segment_info` or copy it?

    segment_info = self.segment_info.copy()
    segment_info['output'] = kwargs.get('matcher_info')
    ....

ZyX-I: @S0lll0s Why? Copying of `self.segment_info` is performed by `Renderer.get_segment_info`. So does merging.

ZyX-I: @S0lll0s So if you will write copying and merging here, it will be repeated again.

S0lll0s: but if I do `super(...).render(side='left', segment_info={'output': kwargs.get('matcher_info')}` then all the other segment info is lost, no?

S0lll0s: ah, no, the super render does a `get_segment_info`. I see.

S0lll0s: I'll do `kwargs['segment_info'] = {'output': kwargs.get('matcher_info')}` instead though, otherwise it will error if someone calls LemonbarRenderer.render(segment_info={})

ZyX-I: @S0lll0s This error is one of the reasons I suggested this solution. `segment_info` supplied on the command-line should not be modified, if you think that passing `segment_info` to renderer is useful, then you need to use at least `kwargs.pop('segment_info', {}).copy()` (note: `pop` and `copy`, your solution will just as well error out).

S0lll0s: ah, you are right. I'm not using it right now, but I also don't know if it may be useful at times. Do you think I should include it or not?

ZyX-I: If it is useful, it will be easy to fix later.

powerline/lint/__init__.py
```diff
@@ -177,8 +183,7 @@ def load_json_config(config_file_path, load=load, open_file=open_file):
 		color_spec(),
 	).optional(),
 	groups=Spec().unknown_spec(
-		group_name_spec().func(check_translated_group_name),
-		group_spec(),
+		group_name_spec().func(check_translated_group_name), group_spec(),
```
ZyX-I: Revert

docs/source/configuration/reference.rst
```diff
@@ -187,6 +187,9 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
     ``out`` and ``rewrite`` prompts (refer to IPython documentation for more 
     details) while ``in`` prompt is the default.
 
+    For wm it is a dictionary ``{output : theme_name}`` that maps the
```
ZyX-I: Remembered that powerline has more then one binding that uses wm: also qtile and awesome. So “For wm (:ref:\`lemonbar <lemonbar-usage>\` only)”.

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,31 +2,27 @@
...
 if __name__ == '__main__':
-	parser = ArgumentParser(description='Powerline BAR bindings.')
+	parser = ArgumentParser(description='Powerline lemonbar bindings.')
 	parser.add_argument(
 		'--i3', action='store_true',
 		help='Subscribe for i3 events.'
 	)
 	args = parser.parse_args()
-	powerline = BarPowerline()
+	powerline = LemonbarPowerline(theme=args.theme)
```
ZyX-I: AttributeError

docs/source/configuration/reference.rst
S0lll0s: *lemonbar only* or *currently lemonbar only*... I am sure this is possible in awesome and probably qtile aswell

ZyX-I: lemonbar only. If something changes documentation in any case needs to be changed, no matter whether currently is there or not.

docs/source/configuration/reference.rst
```diff
@@ -187,6 +187,9 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
     ``out`` and ``rewrite`` prompts (refer to IPython documentation for more 
     details) while ``in`` prompt is the default.
 
+    For wm (:ref:`lemonbar <lemonbar-usage>` only) it is a dictionary ``{output : theme_name}`` that maps the
```
ZyX-I: This line is too long now.

S0lll0s:
the travis error is unrelated: `OSError: [Errno 28] No space left on device`

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
...
+	powerline.update_renderer()
+	bars = []
+	active_screens = [match.groupdict() for match in re.finditer(
+		'^(?P<name>\w+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
+		run_cmd(powerline.pl, ['xrandr', '-q']),
+		re.MULTILINE
+	)]
+
+	for screen in active_screens:
+		command = ['bar-aint-recursive', '-g', '{}x{}+{}'.format(screen['width'], args.height, screen['x'])] + args.args[1:]
```
ZyX-I: Wondering where is output selection.

S0lll0s: what do you mean output selection?

powerline/bindings/bar/powerline-bar.py
```diff
@@ -2,31 +2,27 @@
...
+	parser = ArgumentParser(description='Powerline lemonbar bindings.')
 	parser.add_argument(
 		'--i3', action='store_true',
 		help='Subscribe for i3 events.'
 	)
 	args = parser.parse_args()
-	powerline = BarPowerline()
+	powerline = LemonbarPowerline()
+	powerline.update_renderer()
+	powerline.pl.warn("the 'bar' bindings are deprecated, please switch to 'lemonbar'")
```
ZyX-I: Message should start with capital letter.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
```
ZyX-I: Missing newline.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
...
+import sys
```
ZyX-I: `sys` is unused.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
...
+import time
+import re
+import subprocess
+
+from threading import Lock, Timer
+from argparse import ArgumentParser, REMAINDER
+
+from powerline.lemonbar import LemonbarPowerline
+from powerline.lib.monotonic import monotonic
```
ZyX-I: monotonic is unused

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
...
+from powerline.lib.dict import mergedicts
```
ZyX-I: mergedicts is unused

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,87 @@
...
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
+	)
+	args = parser.parse_args()
+
+	powerline = LemonbarPowerline()
+	powerline.update_renderer()
+	bars = []
+	active_screens = [match.groupdict() for match in re.finditer(
+		'^(?P<name>\w+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
```
ZyX-I: How is `\w+` supposed to match what you said in linter is `^[A-Z0-9-]+$`?

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: Also you must not use `{}`. This is not supported on Python-2.6, use explicit argument numbers.

ZyX-I: @S0lll0s How is lemonbar supposed to know what monitor to display its bar on?

ZyX-I:
PR will not be accepted without working functional tests from S0lll0s/powerline#2.

S0lll0s:
I'll update the documentation quickly now

S0lll0s:
Should I squash the documentation change commit into the `deprecate bar-aint-recursive` commit or should i seperate the doc changes from that commit and put them all in one seperate?

ZyX-I:
07.01.2016, 05:31, "Sol Bekic" <notifications@github.com>:
> Should I squash the documentation change commit into the deprecate bar-aint-recursive commit or should i seperate the doc changes from that commit and put them all in one seperate?

What you think is best. But do not loose my commits in process, I do not see them.

docs/source/usage/wm-widgets.rst
```diff
@@ -50,24 +50,27 @@ Add the following to :file:`~/.config/qtile/config.py`:
...
 
-To run the bar simply pipe the output of the binding script into ``bar`` and 
-specify appropriate options, for example like this::
+To run the bar simply start the binding script:
 
-    python /path/to/powerline/bindings/bar/powerline-bar.py | bar
+    python /path/to/powerline/bindings/lemonbar/powerline-lemonbar.py
 
-to run with i3, simply ``exec`` this in i3 config file::
+You can specify options to be passed to ``lemonbar`` after ``--``, like so:
```
ZyX-I: You are not using `lemonbar`, scripts contains `bar-aint-recursive`.

ZyX-I: (BTW, since there are such problems with naming maybe there should be additional argument like `--command=lemonbar`? I doubt that all systems which have lemonbar have both `lemonbar` and `bar-aint-recursive` programs (symlinks?).)

ZyX-I: Ah, you have already modified this and also added the argument in question.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,92 @@
...
+		help='Refresh interval'
+	)
+	parser.add_argument(
+		'args', nargs=REMAINDER,
+		help='Extra arguments for lemonbar'
+	)
+	parser.add_argument(
+		'--bar-command', '-c',
+		default='lemonbar',
+		help='Name of the lemonbar executable to use'
```
ZyX-I: Missing dot.

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: And you have lost all other my changes regarding dots and linter errors in this file.

powerline/bindings/lemonbar/powerline-lemonbar.py
ZyX-I: Note that this should also be tested.

ZyX-I: And this should be placed *before* REMAINDER argument. I am not sure whether it is supposed to work in this position at all or not.

S0lll0s: actually it does work even after the remainder, but I will move it for clarity anyway.

tests/run_bar_tests.sh
```diff
@@ -0,0 +1,199 @@
...
+			height="${height#=}"
+			height="${height%% *}"
+		fi
+		command="lemonbar"
+		if test "x${args#-c}" != "x$args" ; then
+			command="${args#-c}"
+			command="${command# }"
+			command="${command#=}"
+			command="${command%% *}"
+		fi
```
S0lll0s: or do you want me to try both `-c` and `--bar-command`?

S0lll0s:
@ZyX-I the warning test is still failing, did it ever work?

ZyX-I:
@S0lll0s It worked locally. Though you need to change test: I again forgot that `dash` does not support `&>` redirection.

S0lll0s:
so instead `2>&1 >` ?

ZyX-I:
@S0lll0s AFAIR the order is reversed. I.e. first you bind stdout to a file, then you bind stderr to stdout, so stderr appears binded to the file too. `2>&1 >` is binding stderr to a stdout, then stdout to a file which means that stdout will be redirected to a file and stderr will be redirected to stdout.

tests/run_bar_tests.sh
ZyX-I: No, it is OK to try only `--bar-command` (*not* `-C`, long versions of arguments are always preferred and if you want to test only one variant test only long). But rename `-c` to `-C` in case somebody will actually add overrides (`-c` stands for `--config-override` in `powerline`/`powerline-render`).

tests/test_bar/path/bar-aint-recursive
```diff
@@ -0,0 +1 @@
+lemonbar
```
ZyX-I: It is better to place `ln -s` call in `run_bar_tests.sh`, and put file name into a variable.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,90 @@
...
+	powerline.update_renderer()
+	bars = []
+	active_screens = [match.groupdict() for match in re.finditer(
+		'^(?P<name>[0-9A-Z-]+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
+		run_cmd(powerline.pl, ['xrandr', '-q']),
+		re.MULTILINE
+	)]
+
+	for screen in active_screens:
+		command = [args.bar_command, '-g', '{}x{}+{}'.format(screen['width'], args.height, screen['x'])] + args.args[1:]
```
ZyX-I: You have lost my change somewhere. `{}` are not accepted in python-2.6, use explicit numbers.

powerline/bindings/lemonbar/powerline-lemonbar.py
```diff
@@ -0,0 +1,90 @@
...
+from argparse import ArgumentParser, REMAINDER
+
+from powerline.lemonbar import LemonbarPowerline
+from powerline.lib.shell import run_cmd
+
+
+if __name__ == '__main__':
+	parser = ArgumentParser(
+		description='Powerline BAR bindings.',
+		usage='%(prog)s [-h] [--i3] [--height HEIGHT] [--interval INTERVAL] [--bar-command CMD] -- args')
```
ZyX-I: (And I still want to know why you are not letting argparse construct the usage message itself.)

S0lll0s: it doesn't really display the `-- args` part well, it looks like this:

    usage: powerline-lemonbar.py [-h] [--i3] [--height HEIGHT]
                                 [--interval INTERVAL] [--bar-command BAR_COMMAND]
                                 ...

    Powerline BAR bindings.

    positional arguments:
      args                  Extra arguments for lemonbar.

but the `--` is mandatory and it doesn't show that.

ZyX-I: @S0lll0s It is not mandatory. But without `--` you can only supply arguments that do not start with hyphen-minus.

S0lll0s: yes, but all `lemonbar` options start with it:

    NAME
           bar - bar ain't recursive

    SYNOPSIS
           bar [-h | -g widthxheight+x+y | -b | -d | -f font | -p | -u pixel | -B color | -F color]

tests/test_configuration.py
S0lll0s: I remember now why I removed it; lemonbar doesn't actually support escapeing that way, whenever it finds a `%{` it matches it, two `%%` don't help. The only thing you need to escape therefore is an `%{`, and you escape it by inserting `%{}` inbetween the two characters to form `%%{}{`.

ZyX-I:
@S0lll0s Can you pull another two commits from the PR? It is regarding the documentation.

ZyX-I:
The resulting documentation looks like

Documentation

ZyX-I:
(Actually I have changed ARGS doc wording, but you get the idea.)

S0lll0s:
you accidentally changed the `--bar-command` shorthand to `-C` which breaks the tests; i reverted that

ZyX-I:
@S0lll0s I *intentionally* changed shorthand to `-C` which is confirmed by a error message. As I said in https://github.com/powerline/powerline/pull/1431/files#r49143228 tests should either test *only* `--bar-command` or both `--bar-command` and `-C`, but not only `-C`.

S0lll0s:
oh, I missed that comment because the diff was outdated again. What happened to all the tests? Did you change something with the encoding? It seems all spaces changed from `0x0a` to `0x2c` (or the other way around?)

ZyX-I:
@S0lll0s Do you use web interface to check this? If yes, do not. In my PR HEAD I see non-breaking spaces, so if somebody did this change it was you.

ZyX-I:
@S0lll0s I actually now see no differences between my PR HEAD and this PR HEAD.

S0lll0s:
@ZyX-I no, I `xxd`'d the errors of my local test run. It seems my vim fucked up the encoding somehow? It file is set to utf-8 though

ZyX-I:
@S0lll0s No, actually you have written `escape` function so it does not replace spaces with non-breaking spaces. I am wondering whether it will do any harm if you escape *only* per cent sign as `%%{}`: i.e. restore `character_translations`, but translate to `%%{}` in place of `%%`. `lemonbar` renderer appears to be the *only* one redefining `escape` function.

S0lll0s:
@ZyX-I a few more characters overall but it shouldn't matter

tests/run_bar_tests.sh
```diff
@@ -0,0 +1,200 @@
...
+	local line="$(head -n1 "$log_file")"
+	local linenum="$(cat "$log_file" | wc -l)"
+	if test $linenum -lt 5 ; then
+		fail "log:lt" F "Script was run not enough times"
+		return 1
+	elif test $linenum -gt 15 ; then
+		fail "log:gt" E "Script was run too many times"
+		return 1
+	fi
+  local expline="%{l}%{F#ffd0d0d0}%{B#ff303030} $text-left %{F-B--u}%{F#ff303030} %{F-B--u}%{r}%{F#ff303030} %{F-B--u}%{F#ffd0d0d0}%{B#ff303030} $text-right %{F-B--u}"
```
ZyX-I: Spaces.

ZyX-I: I mean, in indentation. Though with most recent suggestion this line should not be different from line in af25681a8003e364e98f11717e62426abc184480.

S0lll0s:
looks like I got all the build errors finally

ZyX-I:
@S0lll0s I guess you need to apply

```Patch
diff --git a/tests/run_bar_tests.sh b/tests/run_bar_tests.sh
index 73a46cc..473ee33 100755
--- a/tests/run_bar_tests.sh
+++ b/tests/run_bar_tests.sh
@@ -31,6 +31,7 @@ run() {
 		XDG_CONFIG_DIRS="$TEST_ROOT/dummy" \
 		PYTHONPATH="$PYTHONPATH" \
 		TEST_ROOT="$TEST_ROOT" \
+		LD_LIBRARY_PATH="$LD_LIBRARY_PATH" \
 		"$@" || true
 }
 
```
to https://github.com/S0lll0s/powerline/commit/e1ad85356be4e3b605f21b05edfe3046de10d531: self-compiled Pythons’ builds are failing.

S0lll0s:
@ZyX-I yep, that made the `UCS2` builds pass aswell.

S0lll0s:
Once again, thanks for the patience @ZyX-I :)

PR #1451, opened 2015-09-14
Show conda environment in virtualenv segment.

QuLogic:
Conda environments are pretty much the same as virtualenvs, so I just added a check for them there. One could put it in a separate segment, but I think it fits naturally here.

### DIFF
--- a/powerline/segments/common/env.py
+++ b/powerline/segments/common/env.py
@@ -19,9 +19,20 @@ def environment(pl, segment_info, variable=None):
 
 
 @requires_segment_info
-def virtualenv(pl, segment_info):
-	'''Return the name of the current Python virtualenv.'''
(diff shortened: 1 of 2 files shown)


ZyX-I:
You are missing tests. And there should be argument(s) that allow ignoring virtualenv or conda environment.

Note: test failure is not related to missing tests.

QuLogic:
Oops, I totally forgot about tests. I've added them, but I wasn't sure how to run them (it's not in `CONTRIBUTING.md`.)

QuLogic:
Never mind, I see `py.test` works well enough. Just need to figure out how to set two environment variables at a time now.

QuLogic:
I added some tests; I assume the failure is the same as before.

PR #1452, opened 2015-09-23
Add git merge/rebase to git branch segment.

russellb:
Prior to using powerline with bash, I used $(__git_ps1) to add git
branch info to my prompt.  __git_ps1 includes the branch name, but it
also provides some additional status information that is incredibly
helpful for staying oriented when in the middle of some git operations
like a merge or rebase.

This patch implements something similar for the git implementation of
the vcs branch segment.  For example, if you're in the middle of an
interactive rebase on branch "master", the branch segment will now
show "master (REBASE-i)".

### DIFF
--- a/powerline/lib/vcs/git.py
+++ b/powerline/lib/vcs/git.py
@@ -16,15 +16,46 @@ _ref_pat = re.compile(br'ref:\s*refs/heads/(.+)')
 
 
 def branch_name_from_config_file(directory, config_file):
+	status = ''
+	b = None
+	if os.path.exists('%s/rebase-apply' % directory):
+		if os.path.exists('%s/rebase-apply/rebasing' % directory):
+			status = ' (REBASE)'
+			with open('%s/rebase-apply/head-name' % directory, 'r') as f:
+				b = f.read()
+		elif os.path.exists('%s/rebase-apply/applying' % directory):
+			status = ' (AM)'
+		else:
+			status = ' (REBASE|AM)'
+	elif os.path.exists('%s/rebase-merge' % directory):
+		with open('%s/rebase-merge/head-name' % directory, 'r') as f:
+			b = f.read()
+		if os.path.exists('%s/rebase-merge/interactive' % directory):
+			status = ' (REBASE-i)'
+		else:
+			status = ' (REBASE-m)'
+	elif os.path.exists('%s/MERGE_HEAD' % directory):
+		status = ' (MERGING)'
+	elif os.path.exists('%s/CHERRY_PICK_HEAD' % directory):
+		status = ' (CHERRY-PICKING)'
+	elif os.path.exists('%s/REVERT_HEAD' % directory):
+		status = ' (REVERT_HEAD)'
+	elif os.path.exists('%s/BISECT_LOG' % directory):
+		status = ' (BISECTING)'
+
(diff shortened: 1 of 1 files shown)


ZyX-I:
1. Missing tests.
2. Should be a separate segment, not part of the branch segment.
3. You *must not* use `'%s/foo' % dir` to concatenate paths. Use `powerline.lib.path.join`. This is also why existing tests are failing.

powerline/lib/vcs/git.py
```diff
@@ -16,15 +16,46 @@
 
 
 def branch_name_from_config_file(directory, config_file):
+	status = ''
+	b = None
+	if os.path.exists('%s/rebase-apply' % directory):
```
ZyX-I: This and next should be `os.path.isdir`, not `os.path.exists`.

russellb:
Thanks for the feedback.  I'll address it.

Regarding a separate segment vs. the branch segment, I debated that and went this route as a first cut.  Currently, say you're in the middle of an interactive rebase on your "master" branch, while editing a commit, the branch segment will just have a random hash.  Now it says "master (REBASE-i)" which I think is a bit more helpful.  If it's a new segment, the branch segment would go back to being a hash.  For the new segment, do you think "master (REBASE-i)" is appropriate?  Just "REBASE-i" ?  Basically, would you think of it as a new type of branch segment specific to git or a new "exttra vcs status" segment?  I suppose another thing could be to make this optional behavior to the branch segment.  Anyway, let me know what you'd like to see a new segment look like.

ZyX-I:
@russellb For the new segment it should be `status_chars['rebase:interactive']`. It is better if branch segment shows `{branch}*` (`*` *may* be a separate subsegment, but emitted by branch segment; not necessary). Do not forget that `branch_name_from_config_file` is supposed to be called *only* when `HEAD` file changes (tracked by inotify).

New VCS segment should not be VCS-specific. But it is not necessary to provide any implementation, *except* the one returning `None` at the end for other VCS types if you don’t know how to get such statuses.

Note about `status_chars['rebase:interactive']`: ` (REBASE-i)` is rather lengthy, but alternatives may be less readable. So actual string needs to be configurable so that it is possible to replace them with unicode characters or less lengthy strings. `:` as a separator is not necessary, but this is a convention for highlight groups, and I expect such segment to return something like

    status_components = status.split(':')
    return [{
        'contents': status_chars[status],
        'highlight_groups': ['dirstatus:' + ':'.join(status_components[:i]) for i in range(len(status_components), 0, -1)] + ['dirstatus']
    }]
(i.e. so that for `rebase:interactive` it will request `['dirstatus:rebase:interactive', 'dirstatus:rebase', 'dirstatus']` highlight groups). Different highlight groups are needed to put more visual pressure on such statuses if needed because they require resolution.

---

About `powerline.lib.path.join`: this function exists because some functions may receive *both* `b'foo/bar'` and `u'foo/bar'` paths (i.e. both bytes and unicode) and it is easier to write such function then to use only one kind of paths, especially given that I must not assume filesystem using specific encoding.

russellb:
Thanks for the detailed feedback!

I haven't had a chance to come back to this so I'm just going to close it out for now.  I'll reopen this if/when I get back to it.

Thanks again.

PR #1453, opened 2015-09-24
Adding note about python-dev requirement for pyuv

dragon788:


ZyX-I:
This is distribution-specific. In Gentoo you need no python-dev (and there is no such a thing). Mac OS X also has no such package.

PR #1461, opened 2015-10-04
Make date segments work with non-ascii characters in the format string.

haakenlid:
Enables fancy glyphs in the date segment
```json
"powerline.segments.common.time.date": { 
    "args": { "format": "%F ⌚ %T" } 
} 
```
image

This solves a bug where unicode characters in the format string would cause a UnicodeEncodeError.
```
UnicodeEncodeError: 'ascii' codec can't encode character
u'\u231a' in position 3: ordinal not in range(128)
```

### DIFF
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -17,7 +17,7 @@ def date(pl, format='%Y-%m-%d', istime=False):
 	Highlight groups used: ``time`` or ``date``.
 	'''
 	return [{
-		'contents': datetime.now().strftime(format),
+		'contents': datetime.now().strftime(format.encode('utf-8')).decode('utf-8'),
 		'highlight_groups': (['time'] if istime else []) + ['date'],
 		'divider_highlight_group': 'time:divider' if istime else None,
 	}]
@@ -33,7 +33,7 @@ def fuzzy_time(pl, unicode_text=False):
 	'''Display the current time as fuzzy time, e.g. "quarter past six".
 
 	:param bool unicode_text:
-		If true then hyphenminuses (regular ASCII ``-``) and single quotes are 
+		If true then hyphenminuses (regular ASCII ``-``) and single quotes are
 		replaced with unicode dashes and apostrophes.
 	'''
 	hour_str = ['twelve', 'one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine', 'ten', 'eleven']
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -19,7 +19,7 @@ def date(pl, format='%Y-%m-%d', istime=False):
 	try:
 		contents = datetime.now().strftime(format)
 	except UnicodeEncodeError:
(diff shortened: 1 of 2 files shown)


powerline/segments/common/time.py
```diff
@@ -33,7 +38,7 @@ def fuzzy_time(pl, unicode_text=False):
 	'''Display the current time as fuzzy time, e.g. "quarter past six".
 
 	:param bool unicode_text:
-		If true then hyphenminuses (regular ASCII ``-``) and single quotes are 
+		If true then hyphenminuses (regular ASCII ``-``) and single quotes are
```
ZyX-I: Whitespace change must not be present.

powerline/segments/common/time.py
```diff
@@ -16,8 +16,13 @@ def date(pl, format='%Y-%m-%d', istime=False):
 
 	Highlight groups used: ``time`` or ``date``.
 	'''
+	try:
+		contents=datetime.now().strftime(format)
```
ZyX-I: Missing spaces around `=`.

powerline/segments/common/time.py
```diff
@@ -16,8 +16,13 @@ def date(pl, format='%Y-%m-%d', istime=False):
...
+	except UnicodeEncodeError:
+		contents=datetime.now().strftime(format.encode('utf-8'))
```
ZyX-I: Missing spaces around `=`.

ZyX-I:
Also missing test which fails without this change on Python-2.

haakenlid:
There's a unit test now.

tests/test_segments.py
```diff
@@ -709,6 +709,7 @@ def test_date(self):
 		with replace_attr(self.module, 'datetime', Args(now=lambda: Args(strftime=lambda fmt: fmt))):
 			self.assertEqual(self.module.date(pl=pl), [{'contents': '%Y-%m-%d', 'highlight_groups': ['date'], 'divider_highlight_group': None}])
 			self.assertEqual(self.module.date(pl=pl, format='%H:%M', istime=True), [{'contents': '%H:%M', 'highlight_groups': ['time', 'date'], 'divider_highlight_group': 'time:divider'}])
+		self.module.date(pl=pl, format='\u231a')
```
ZyX-I: This needs to be `self.assertEqual`. It is possible that such change will require using `contents = datetime.now().strftime(format.encode('utf-8')).decode('utf-8')` in `time.py` in place of what is there now as I am not sure that `u'\u231a'` can be compared with string which `date` will return without some unicode error.

haakenlid:
Changed the test case.

tests/test_segments.py
haakenlid: I've changed the test case to be similar to the rest.

ZyX-I:
This test hits a bug in PyPy3. Bad I did not mention it before merging in a839b15ecd7d8fde03b73c672c1798f92c42b8fc.

ZyX-I:
https://bitbucket.org/pypy/pypy/issues/2161/pypy3-strftime-does-not-accept-unicode

PR #1463, opened 2015-10-05
Missed a letter

btroo:


ZyX-I:
Commit message is not imperative.

Issue #1281, opened 2015-01-22
Powerline causes flickering in vim.


person808:
After opening a file with vim I can scroll flicker-free. When I open another file, or even another buffer like Unite, vim starts to flicker when switching files and scrolling. Repeated switching makes the flickering worse. Sometimes, I can't even see text until I stop scrolling.

```
vim --version
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jan 14 2015 20:16:36)
Included patches: 1-580
Compiled by Arch Linux
Huge version with GTK2 GUI.  Features included (+) or not (-):
+acl             -ebcdic          +localmap        +printer         +title
+arabic          +emacs_tags      +lua             +profile         +toolbar
+autocmd         +eval            +menu            +python          +user_commands
+balloon_eval    +ex_extra        +mksession       -python3         +vertsplit
+browse          +extra_search    +modify_fname    +quickfix        +virtualedit
++builtin_terms  +farsi           +mouse           +reltime         +visual
+byte_offset     +file_in_path    +mouseshape      +rightleft       +visualextra
+cindent         +find_in_path    +mouse_dec       +ruby            +viminfo
+clientserver    +float           +mouse_gpm       +scrollbind      +vreplace
+clipboard       +folding         -mouse_jsbterm   +signs           +wildignore
+cmdline_compl   -footer          +mouse_netterm   +smartindent     +wildmenu
+cmdline_hist    +fork()          +mouse_sgr       -sniff           +windows
+cmdline_info    +gettext         -mouse_sysmouse  +startuptime     +writebackup
+comments        -hangul_input    +mouse_urxvt     +statusline      +X11
+conceal         +iconv           +mouse_xterm     -sun_workshop    -xfontset
+cryptv          +insert_expand   +multi_byte      +syntax          +xim
+cscope          +jumplist        +multi_lang      +tag_binary      +xsmp_interact
+cursorbind      +keymap          -mzscheme        +tag_old_static  +xterm_clipboard
+cursorshape     +langmap         +netbeans_intg   -tag_any_white   -xterm_save
+dialog_con_gui  +libcall         +path_extra      -tcl             -xpm
+diff            +linebreak       +perl            +terminfo
+digraphs        +lispindent      +persistent_undo +termresponse
+dnd             +listcmds        +postscript      +textobjects
```

Terminal is gnome terminal.

ZyX-I:
What is your configuration? Are some of those files version-controlled?

You can use profiling to get the results:

[...]

PR #1491, opened 2015-11-14
Tabline and bufferlister updates

johnmiked15:
This pull request dramatically improves the experience of using the tabline with `showtabline=2` set. #1281 discusses an issue with Powerline flickering that is only caused when the bufferline is always displayed (with only one tab).

This is due to the prior usage of `vim_getbufoption(segment, 'buflisted')` in the conditionals of the bufferlister segment. Querying the buffer options through the vim python module's `option` attribute caused vim to think it needed to update the tabline for every keystroke after any event that changed the buffer's options.

Using the vim module's `eval` method to directly use the `buflisted(nr)` vim method instead does not cause vim to update the tabline after every keystroke, but rather after events that would change that status, thus fixing the flickering problem.

In addition to improving the flickering, I have also added additional highlight groups to further customize the colors used by the tabline when buffers are listed. You can now customize the colors based on a buffer's modified status, and the tabline has its own background and divider highlight group by default.

### DIFF
--- a/powerline/config_files/colorschemes/vim/__main__.json
+++ b/powerline/config_files/colorschemes/vim/__main__.json
@@ -23,15 +23,24 @@
 		"csv:column_number": "line_current",
 		"csv:column_name":   "line_current_symbol",
 
+		"tab:background":     "background",
+		"tab:divider":        "background:divider",
+
 		"tab_nc:modified_indicator": "modified_indicator",
 		"tab_nc:file_directory":     "information:unimportant",
 		"tab_nc:file_name":          "tab_nc:file_directory",
 		"tab_nc:tabnr":              "tab_nc:file_directory",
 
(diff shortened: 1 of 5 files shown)


ZyX-I:
LGTM, except for two things:

1. Use `str.format()` in place of `str.__mod__` (`str %`).
2. Please add a comment describing why you are not using `vim_getbufoption` into the source code.

johnmiked15:
Just noticed I missed one. I'll get it soon.

ZyX-I:
I actually expected you to merge squash the last commit with the first one.

johnmiked15:
I can do that.

johnmiked15:
Hmm, I missed one usage of the % operator in my original request. Line 121 of `listers/vim.py`. Did you want me to fix that up?

PR #1501, opened 2015-11-23
[Players - mpd] Use Unicode where available

### LINKED ISSUE #1500: [Players - mpd] Exception when track metadata contains non-ASCII text
When using Python 2, if mpd sends back track information containing Unicode, a `UnicodeDecodeError` is raised because it tries to convert the strings to Unicode assuming they're ASCII.

Will post PR in a minute.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28527288-players-mpd-exception-when-track-metadata-contains-non-ascii-text?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mook:
Otherwise the mpd function will fail when it encounters non-ASCII metadata as it tries to do a unicode.format() and attempts to decode the incoming data as ASCII, throwing a UnicodeDecodeError exception.

Fixes #1500 

Note that `python-mpd` (as opposed to `python-mpd2`) does not support `client.currentsong()`, so I don't think that actually works anymore.  Otherwise I'd be happy to a `try / except TypeError` + redo without the `use_unicode` parameter.  `python-mpd2` is happy with the extra argument under Python 3 (it's just mostly a no-op).

### DIFF
--- a/powerline/segments/common/players.py
+++ b/powerline/segments/common/players.py
@@ -189,7 +189,11 @@ class MpdPlayerSegment(PlayerSegment):
 				'total': now_playing[3],
 			}
 		else:
-			client = mpd.MPDClient()
+			try:
+				client = mpd.MPDClient(use_unicode=True)
+			except TypeError:
+				# python-mpd 1.x does not support use_unicode
+				client = mpd.MPDClient()
 			client.connect(host, port)
 			if password:
 				client.password(password)


ZyX-I:
`python-mpd` *does* support `currentsong`, but not `use_unicode`.

mook:
Sorry about that; my testing was bad.  Patch fixed to try `use_unicode` and fall back instead; hopefully that's not too terrible.

PR #1508, opened 2015-12-09
Add support for OS X en0 on internal_ip

kamushadenes:
This simple change ensures support for OS X interfaces on the common.net.internal_ip segment.

### DIFF
--- a/powerline/segments/common/net.py
+++ b/powerline/segments/common/net.py
@@ -74,7 +74,7 @@ else:
 	_interface_starts = {
 		'eth':      10,  # Regular ethernet adapters         : eth1
 		'enp':      10,  # Regular ethernet adapters, Gentoo : enp2s0
-        'en':       10,  # OS X                              : en0 
+		'en':       10,  # OS X                              : en0 
 		'ath':       9,  # Atheros WiFi adapters             : ath0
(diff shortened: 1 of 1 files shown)


powerline/segments/common/net.py
```diff
@@ -74,6 +74,7 @@ def internal_ip(pl, interface='auto', ipv=4):
 	_interface_starts = {
 		'eth':      10,  # Regular ethernet adapters         : eth1
 		'enp':      10,  # Regular ethernet adapters, Gentoo : enp2s0
+        'en':       10,  # OS X                              : en0
```
ZyX-I: 10 looks misaligned. It is really misaligned or this is GH problem?

kamushadenes: Wondered the same when I made the pull request, but on vim looks ok (expandtabs deactivated), and so does in Sublime.

Anyway, it's inside a dict definition, so identation shouldn't really matter, and it's working on my system.

ZyX-I: @kamushadenes I have checked and you are using spaces. You must change this to tabs. Alignment is OK, but indentation is wrong.

kamushadenes: Indeed. Weird... well, fixed.

Issue #1512, opened 2016-01-01
External IP segment causes entire status line to fail if unable to connect


bacondropped:
If Internet connection is down, or powerline is unable to resolve given URL, or any other network-based exception occurs during the call of `_external_ip(query_url)` in `segments/common/net.py`, that function raises an unhandled exception which causes entire powerline to display that exception.

From the user perspective, displaying a concise error message instead, something like `No connection` or disabling the segment altogether would provide a smoother user experience. I'm submitting a pull request which implements this behavior.

ZyX-I:
Where and what exactly do you see? Exceptions are already caught and logged, but if you are not redirecting log (there is configuration option for this) and are not using daemon you will see the exception in tmux.

ZyX-I:
I have added `raise ValueError` to the `_external_ip` function and the resulting exception is caught and logged as expected.

PR #1513, opened 2016-01-01
Disable ExternalIpSegment on network exceptions

bacondropped:
If  external  network  is  unavailable, ExternalIpSegment  will  fail  with  an
exception and  cause powerline to display  error text instead of  all segments.
This commit fixes  it by catching all exceptions  inside _external_ip function,
returning an empty  string in case of  an error which causes  Powerline to hide
ExternalIpSegment.

This pull request is referenced in issue #1512.

### DIFF
--- a/powerline/segments/common/net.py
+++ b/powerline/segments/common/net.py
@@ -30,7 +30,10 @@ def hostname(pl, segment_info, only_if_ssh=False, exclude_domain=False):
 
 
 def _external_ip(query_url='http://ipv4.icanhazip.com/'):
-	return urllib_read(query_url).strip()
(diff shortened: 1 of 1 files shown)


ZyX-I:
If exception is not caught by higher-level catch-all try-except block (present in ThreadedSegment) then this is ThreadedSegment bug. In any case I will never accept code that

1. Uses `except:` clause to silence an exception. There is logging facility.
2. Uses `""` to disable segment. There is `None` for such returns.
3. It also does *not* return an empty string in case of exception. It *disables* segment due to `if not ip` condition.

PR #1517, opened 2016-01-12
Add listers for i3wm workspaces and outputs

S0lll0s:


S0lll0s:
Also (especially during development) a `seginfo` segment could be useful that returns a `segment_info`  value based on a specifyable key in `"args"`.

S0lll0s:
damn, rebasing went wrong

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+
+conn = None
+
+
+@requires_segment_info
+def output_lister(pl, segment_info):
+	'''List all outputs in segment_info format'''
+
+	outputs = re.findall(
+		'^([0-9A-Za-z-]+) connected',
```
ZyX-I: Do not use `re.findall` here, use `re.compile` and operate on resulting object. `re.*` methods are only acceptable when they run once on one regex.

I would also highly suggest this to abstract away into a function used by both this file and script.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+
+
+conn = None
+
+
+@requires_segment_info
+def output_lister(pl, segment_info):
```
S0lll0s: recommenting bc diff is outdated now: this lister isn't i3-specific at all and I could imagine someone implementing output support for awesome or Qtile some time (maybe I'll do it myself for awesome when I find time and on my next setup) - should it be moved to a `listers/wm.py` file? Or should that only happen when another renderer or binding actually supports outputs @ZyX-I

powerline/listers/i3wm.py
S0lll0s: >I would also highly suggest this to abstract away into a function used by both this file and script.

what script? I thought `re.findall` was "smart enough" to do that internally but I will do that.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+import re
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+
+
+conn = None
+
+
+@requires_segment_info
```
ZyX-I: I do not see `segment_info` dictionary used anywhere here. But it should be used: specifically you should create a copy of `segment_info` and put output there like other listers do, *not* return a dictionary with just output.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+			{
+				'output': output
+			},
+			{}
+		)
+		for output in outputs
+	)
+
+
+@requires_segment_info
```
ZyX-I: Same applies here.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+	output = output or segment_info.get('output')
+
+	return (
+		(
+			{
+				'workspace': w['name'],
+				'output': w['output'],
+				'visible': w['visible'],
+				'urgent': w['urgent'],
+				'focused': w['focused']
```
ZyX-I: I do not see any segments actually using this information.

powerline/listers/i3wm.py
S0lll0s: ah, I was wondering about that, but `listers.pdb.frame_lister` doesn't either:

```python
@requires_segment_info
def frame_lister(pl, segment_info, full_stack=False, maxframes=3):
	'''List all frames in segment_info format

	:param bool full_stack:
		If true, then all frames in the stack are listed. Normally N first 
		frames are discarded where N is a number of frames present at the first 
		invocation of the prompt minus one.
	:param int maxframes:
		Maximum number of frames to display.
	'''
	if full_stack:
		initial_stack_length = 0
		frames = segment_info['pdb'].stack
	else:
		initial_stack_length = segment_info['initial_stack_length']
		frames = segment_info['pdb'].stack[initial_stack_length:]

	if len(frames) > maxframes:
		frames = frames[-maxframes:]

	return (
		(
			{
				'curframe': frame[0],
				'initial_stack_length': initial_stack_length,
			},
			{}
		)
		for frame in frames
	)
```

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,78 @@
...
+			},
+			{}
+		)
+		for w in conn.get_workspaces()
+			if (not only_show or any(w[typ] for typ in only_show))
+			and (not output or w['output'] == output)
```
ZyX-I: This looks very strange. `if` should be indented with `for`. `and` should not be on the next line at all *or* the whole `if` conditional should be inside parenthesis.

powerline/listers/i3wm.py
S0lll0s: so one copy per lister result?

powerline/listers/i3wm.py
ZyX-I: Apparently frame_lister contains a bug.

powerline/listers/i3wm.py
ZyX-I: Yes. Better use vim listers as a reference.

powerline/listers/i3wm.py
S0lll0s: copied it from [`segments/i3wm.py:56`](https://github.com/powerline/powerline/blob/develop/powerline/segments/i3wm.py#L56)

powerline/listers/i3wm.py
ZyX-I: Also it would look much better if there was one new `workspace` key. Key `visible` in segment_info looks like the key that tells segment it is (in)visible, not like the key that has something to do with workspace. Also note that lister should list only keys that were already present without the lister. PDB lister is no exception, both keys it lists are initially set by renderer.

ZyX-I: Last demand is because segments are not expected and must not be bound to being used outside of the lister.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,81 @@
...
+	for key in kwargs:
+		copy[key] = kwargs[key]
+	return copy
+
+
+@requires_segment_info
+def output_lister(pl, segment_info):
+	'''List all outputs in segment_info format'''
+
+	regex = re.compile('^([0-9A-Za-z-]+) connected')
```
ZyX-I: Absolutely useless. `re.compile` must be called only *once*, so it *must not* be called inside a function. Use some global like `XRANDR_OUTPUT_RE`.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,81 @@
...
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+
+
+conn = None
+
+
+def update_segment_info(segment_info, **kwargs):
+	copy = segment_info.copy()
+	for key in kwargs:
```
ZyX-I: Do not iterate, use `dict.update`.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,81 @@
...
+import re
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+
+
+conn = None
+
+
+def update_segment_info(segment_info, **kwargs):
```
ZyX-I: Function has absolutely nothing to do with segment_info. Better place this in `powerline.lib.dict`:

```Python


def updated(d, *args, **kwargs):
	'''Copy dictionary and update it with provided arguments
	'''
	d = d.copy()
	d.update(*args, **kwargs)
	return d
```

: I use this function in `editors-refactor` branch.

ZyX-I: (“updated” is named after “sorted”)

powerline/listers/i3wm.py
S0lll0s: you mean *inside*, yeah, I get it. There currently aren't any segments, but they would work similar to how the `i3wm.workspaces` segment works with `output`: filter information based on the workspace specified in `args`, which if not present can be overriden by the lister if it is being used.

powerline/listers/i3wm.py
ZyX-I: I know. This does not make it look right, just I missed this earlier.

powerline/listers/i3wm.py
ZyX-I: Segments should be added before the lister, or user will think lister does something useful while it does not.

powerline/listers/i3wm.py
S0lll0s: not moving to another file though because this regex doesn't match the other (size/position) info and I can't use the "full" regex with `.findall`

ZyX-I: It should not be in the other file, but `re.compile` call should not define any local variable. Things like XRANDR_OUTPUT_RE should normally have the same scope as function they are used in.

ZyX-I: I actually cannot remember any compiled regex used outside of the module it is defined in.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,79 @@
...
+
+import re
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+from powerline.lib.dict import updated
+
+
+conn = None
+XRANDR_OUTPUT_RE = re.compile('^([0-9A-Za-z-]+) connected')
```
ZyX-I: If you meant my comment regarding separate function: I do not see why can’t you just write

```Python
def get_connected_xrandr_outputs(pl):
	return (match.groupdict() for match in re.finditer(
		'^(?P<name>[0-9A-Za-z-]+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)',
		run_cmd(pl, ['xrandr', '-q']),
		re.MULTILINE
	))
```

and in lister simply write `output=output['name']`.

ZyX-I: (Except that regex should be carried away as well: it was fine when it was there in code that is run once.)

S0lll0s: and what file should I put that in?

ZyX-I: I think it is better to keep it here for now. Correct location is something like powerline/bindings/wm.py, but adding new file for such a small single function…

S0lll0s: and then have the binding require it from there? Or duplicate it for now

ZyX-I: Of course, import, that’s the whole point.

S0lll0s:
And I suppose you want tests for this aswell? I haven't found any other lister tests, is there something to reference somewhere?

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+		else:
+			conn = i3ipc.Connection()
+
+	output = output or segment_info.get('output')
+
+	return (
+		(
+			updated(segment_info,
+				output = w['output'],
+				workspace = {
```
ZyX-I: No spaces in arguments. Also this should be aligned with `segment_info` and *not* use tabs for aligment or `segment_info` should be on the next line.

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+		if ((not only_show or any(w[typ] for typ in only_show))
+			and (not output or w['output'] == output))
 	]
 
 
 @requires_segment_info
+def workspace(pl, segment_info, workspace=None, strip=0):
+	'''Return state for the specified workspace
+
+	:param str workspaces:
```
ZyX-I: There is no such parameter.

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+		Specifies which workspace to show. If unspecified, may be set by the ``list_workspaces`` lister.
```
ZyX-I: Too long line.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+					'visible': w['visible'],
+					'urgent': w['urgent'],
+					'focused': w['focused']
+				}
+			),
+			{}
+		)
+		for w in conn.get_workspaces()
+		if (((not only_show or any(w[typ] for typ in only_show))
+			and (not output or w['output'] == output)))
```
ZyX-I: Tabs are *not* used for alignment.

ZyX-I:
@S0lll0s All listers are tested, though implicitly: pdb tests in test_shells and test_tabline.vim.

powerline/listers/i3wm.py
ZyX-I: (And it is usually preferred to write code so that no alignment is needed in first place. I have alignment mostly for function arguments: e.g. check powerline.Powerline.init.)

ZyX-I:
Since listers are not used in configuration then you will have to write explicit test.

docs/source/develop/segments.rst
```diff
@@ -517,6 +517,16 @@ Pdb
...
 
+i3wm
+----
+
+``mode``
+    Currently active i3 mode (as a string).
+
+``output``
+    ``xrandr`` output name currently drawing to. Currently only available
+    in lemonbar bindings.
```
ZyX-I: Missing `workspace` key documentation.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+import re
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+from powerline.lib.dict import updated
+
+
+conn = None
```
ZyX-I: This should be near `workspace_lister`. Though no, since connection is shared between lister and segment it (not it directly, but function that gets connection) should be in `powerline/bindings/wm/__init__.py`, with `get_connected_xrandr_outputs`: these two functions are long enough.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+XRANDR_OUTPUT_RE = re.compile('^(?P<name>[0-9A-Za-z-]+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)', re.MULTILINE)
```
ZyX-I: And this should be two blank lines away from get_connected_xrandr_outputs. I actually would like this close position, but AFAIR linter does not accept this.

(BTW, it is better to write `r'…'`: I do not like relying on `'\d'` being interpreted as `'\\d'`.)

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+def get_connected_xrandr_outputs(pl):
+    return [match.groupdict() for match in XRANDR_OUTPUT_RE.finditer(
```
ZyX-I: Missing doc + this should be a generator expression, not list comprehension.

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
 		'contents': w['name'][min(len(w['name']), strip):],
 		'highlight_groups': calcgrp(w)
 	} for w in conn.get_workspaces()
-		if (not only_show or any(w[typ] for typ in only_show))
-		and (not output or w['output'] == output)
+		if ((not only_show or any(w[typ] for typ in only_show))
+			and (not output or w['output'] == output))
```
ZyX-I: Same: align with spaces.

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+	else:
+		global conn
+		if not conn:
+			try:
+				import i3ipc
+			except ImportError:
+				import i3 as conn
+			else:
+				conn = i3ipc.Connection()
+		w = next(w for w in conn.get_workspaces() if w['name'] == workspace)
```
ZyX-I: `func(expr for args in expr)` is not supported on Python-2.6. You must add additional level of parenthesis to surround generator expression.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+			conn = i3ipc.Connection()
+
+	output = output or segment_info.get('output')
+
+	return (
+		(
+			updated(segment_info,
+				output = w['output'],
+				workspace = {
+					'name': w['name'],
```
ZyX-I: Though given that here you will need sequence “indentation (as `updated`), alignment (as `segment_info` argument), indentation again (nothing to align with)” I would say `segment_info` should be on its own line.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,80 @@
...
+					'visible': w['visible'],
+					'urgent': w['urgent'],
+					'focused': w['focused']
```
ZyX-I: And trailing comma. Here and a line below.

Trailing commas are needed to make adding new key or argument produce *one* line in diff, not *two* (one for new comma, one for new argument).

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
 		'contents': w['name'][min(len(w['name']), strip):],
```
ZyX-I: You have *three* instances of the code that connects to i3. This is too much.

ZyX-I: Also now I got your idea behind workspaces and will say that this segment should thus be deprecated (you want to replace it with lister+workspace segment, right?), see 3c1d5c615ce4c364a9b922fd46e573a312b107a8 (specifically you need to see where and how linter was modified).

powerline/segments/i3wm.py
```diff
@@ -54,12 +54,44 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 	} for w in conn.get_workspaces()
-		if (not only_show or any(w[typ] for typ in only_show))
-		and (not output or w['output'] == output)
+		if ((not only_show or any(w[typ] for typ in only_show))
+			and (not output or w['output'] == output))
 	]
 
 
 @requires_segment_info
+def workspace(pl, segment_info, workspace=None, strip=0):
```
ZyX-I: Do not use `strip`, use `format`. If needed stripping should be done by `str.format`.

ZyX-I: (AFAIR this was previously discussed, do not know why I see `strip` in `workspaces` argument.)

powerline/listers/i3wm.py
S0lll0s: but should i3-specific `conn` handling and  wm-agnostic `xrandr´ handling go in the same file? I guess it doesn't matter so much when the connection isn't made if it isn't used but still

S0lll0s:
I don't get why doing a `git rebase -i HEAD~5` changes commits further back than the only change I do  in the interactive prompt... if I just fixup the second-to-last commit with the last one, how can the hash 10 commits back change?!

powerline/segments/i3wm.py
S0lll0s: and how should stripping like this be done using `str.format`? There is no string splicing support, you can only access single characters of a string. However stripping the first n characters (or even better, stripping leading `[0-9]+: ?` is a __very__ common use-case for this.

ZyX-I: Then absense of first characters stripping support is why `strip` argument was kept in `workspaces`.

ZyX-I: Why you do not have boolean option for stripping `[0-9]+: ` then? Though if workspace names start with `[0-9]+:` always I would suggest strip `.*? ` instead (because this can be done without using any regexes).

S0lll0s: they do not always, but the i3 manual recommends doing it that way. i3 sorts workspaces only if they start with a number like that, but the colon is optional.

>Note that the workspace will hella be named "1: mail". i3 treats workspace names beginning with a number in a slightly special way. Normally, named workspaces are ordered the way they appeared. When they start with a number, i3 will order them numerically. Also, you will be able to use workspace number 1 to switch to the workspace which begins with number 1, regardless of which name it has. This is useful in case you are changing the workspace’s name dynamically. To combine both commands you can use workspace number 1: mail to specify a default name if there’s currently no workspace starting with a "1".

I guess an option that acts like the i3bar options would work though (basically removing `[0-9]+:` if existing, then `str.strip`). What do you mean by strip `.*?`? That would just match everything, no?

ZyX-I: @S0lll0s It would not match everything in any case: without trailing space it will match nothing, but I meant “everything up to first space” (so to use `string.partition(' ')[-1]`). Maybe `string.partition(':')[-1].lstrip()` if omitting space is common.

S0lll0s: @ZyX-I but this will also split workspaces if they contain a colon somewhere else (like `a:bc`), i3bar only splits if its a number followed by a colon at the beginning of the string.

powerline/segments/i3wm.py
```diff
@@ -19,8 +22,14 @@ def calcgrp(w):
...
 
+def format_name(name, strip=False):
+	if strip:
+		return WORKSPACE_REGEX.sub('', name, count=1)
+	return name
+
+
 @requires_segment_info
-def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
+def workspaces(pl, segment_info, only_show=None, output=None, strip_numbers=False):
```
ZyX-I: This change is backwards incompatible and can’t be accepted.

powerline/bindings/wm/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+			import i3 as conn
+		else:
+			conn = i3ipc.Connection()
+	return conn
+
+
+XRANDR_OUTPUT_RE = re.compile(r'^(?P<name>[0-9A-Za-z-]+) connected (?P<width>\d+)x(?P<height>\d+)\+(?P<x>\d+)\+(?P<y>\d+)', re.MULTILINE)
+
+
+def get_connected_xrandr_outputs(pl):
```
ZyX-I: Missing documentation.

powerline/bindings/wm/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+import re
+
+from powerline.theme import requires_segment_info
+from powerline.lib.shell import run_cmd
+
+
+conn = None
+
+
+def get_i3_connection():
```
ZyX-I: Same.

powerline/bindings/wm/__init__.py
```diff
@@ -0,0 +1,31 @@
+# vim:fileencoding=utf-8:noet
```
ZyX-I: I asked for i3 and xrandr in one module because it is `….bindings.wm` and both are wm-related. If there were more functions like this it would make sense to use `….wm` and `….i3`, but currently as i3 ∈ wm they may be in the same file.

powerline/listers/i3wm.py
```diff
@@ -0,0 +1,66 @@
...
+					'focused': w['focused'],
+				},
+			),
+			{
+				'draw_inner_divider': None
+			}
+		)
+		for w in get_i3_connection().get_workspaces()
+		if (((not only_show or any(w[typ] for typ in only_show))
+		   and (not output or w['output'] == output)))
```
ZyX-I: It appears that `and` is aligned incorrectly: should be one column after opening parenthesis.

powerline/segments/i3wm.py
```diff
@@ -33,33 +42,57 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 	return [{
-		'contents': w['name'][min(len(w['name']), strip):],
+		'contents': format_name(w['name'], strip=strip_numbers),
 		'highlight_groups': calcgrp(w)
-	} for w in conn.get_workspaces()
-		if (not only_show or any(w[typ] for typ in only_show))
-		and (not output or w['output'] == output)
+	} for w in get_i3_connection().get_workspaces()
+		if ((not only_show or any(w[typ] for typ in only_show))
+		     and (not output or w['output'] == output))
```
ZyX-I: And this one is one column too far.

S0lll0s:
@ZyX-I again for tests, should I just include a new segment test that runs the `workspace` segment wrapped in the two listers?

ZyX-I:
@S0lll0s Listers and workspace segment should have different tests.

S0lll0s:
@ZyX-I but all just render output tests in `test_segments.py`?

ZyX-I:
@S0lll0s ? In `test_segments.py` it is tested that segments return something specific, when supplied some arguments, renderer is not involved.

S0lll0s:
@ZyX-I nvm, figured that out aswell

S0lll0s:
@ZyX-I Did you take a look at the latest?

powerline/segments/i3wm.py
```diff
@@ -19,8 +22,14 @@ def calcgrp(w):
 	return group
 
 
+def format_name(name, strip=False):
+	if strip:
```
ZyX-I: If strip is integer it should still exhibit the same behaviour. Also for checking type it is recommended to use `isinstance`.

powerline/segments/i3wm.py
```diff
@@ -33,33 +42,64 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+	:param bool strip:
+		If set to True, workspaces prefixed with a number and colon 
+		(like ``1: www`` or ``1:www``) will have only the name shown (``www``).
+		Default: False.
+
+	Highlight groups used: ``workspace`` or ``w_visible``, ``workspace`` or ``w_focused``, ``workspace`` or ``w_urgent``.
+	'''
+
+	if type(strip) == int:
+		pl.warn("integer values for 'strip' are deprecated")
```
ZyX-I: And this warning should be removed, and integer arguments treated as regular boolean values.

powerline/segments/i3wm.py
```diff
@@ -33,33 +42,64 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+
+	workspace = workspace or segment_info.get('workspace')['name']
+	if segment_info.get('workspace') and segment_info.get('workspace')['name'] == workspace:
+		w = segment_info.get('workspace')
+	else:
+		try:
+			w = next((
+				w for w in get_i3_connection().get_workspaces()
+				  if w['name'] == workspace
```
ZyX-I: There should be no alignment.

tests/test_segments.py
```diff
@@ -914,7 +925,7 @@ def test_workspaces(self):
 				{'contents': '3: w3', 'highlight_groups': ['w_urgent', 'w_visible', 'workspace']},
 				{'contents': '4: w4', 'highlight_groups': ['w_focused', 'w_urgent', 'w_visible', 'workspace']},
 			])
-			self.assertEqual(i3wm.workspaces(pl=pl, segment_info=segment_info, only_show=['visible'], strip=3), [
+			self.assertEqual(i3wm.workspaces(pl=pl, segment_info=segment_info, only_show=['visible'], strip=True), [
```
ZyX-I: Revert. Deprecated functionality still needs testing. Also note that `strip=3` needs to be applied to workspace names like `foo` to check that deprecated functionality works as it used to work.

tests/test_segments.py
```diff
@@ -943,6 +974,199 @@ def test_mode(self):
 		self.assertEqual(i3wm.mode(pl=pl, segment_info={'mode': 'default'}, names={'default': 'test'}), 'test')
 		self.assertEqual(i3wm.mode(pl=pl, segment_info={'mode': 'test'}, names={'default': 'test', 'test': 't'}), 't')
 
+	def test_listers(self):
```
ZyX-I: This is better to be placed into its own file, this file is named “test_*segments*”.

tests/test_segments.py
```diff
@@ -883,14 +884,24 @@ def test_weather(self):
...
+	def get_outputs(self, pl):
+		return iter([
+			{'name': 'LVDS1'},
+			{'name': 'HDMI1'},
+			{'name': 'DVI01'},
+		])
+
+	def test_workspaces(self):
+		pl = Pl()
+		with replace_attr(i3wm, "get_i3_connection", lambda: Args(get_workspaces=self.get_workspaces)):
```
ZyX-I: And single quotes.

tests/test_segments.py
```diff
@@ -925,17 +936,37 @@ def test_workspaces(self):
...
 				{'contents': '3: w3', 'highlight_groups': ['w_urgent', 'w_visible', 'workspace']},
 			])
-			self.assertEqual(i3wm.workspaces(pl=pl, segment_info=segment_info, only_show=['visible'], strip=3), [
+			self.assertEqual(i3wm.workspaces(pl=pl, segment_info=segment_info, only_show=['visible'], strip=True), [
 				{'contents': 'w2', 'highlight_groups': ['w_visible', 'workspace']},
 			])
 
+	def test_workspace(self):
+		pl = Pl()
+		with replace_attr(i3wm, "get_i3_connection", lambda: Args(get_workspaces=self.get_workspaces)):
```
ZyX-I: Quotes.

tests/test_segments.py
```diff
@@ -883,14 +884,24 @@ def test_weather(self):
...
+	def get_workspaces(self):
+		return iter([
 			{'name': '1: w1', 'output': 'LVDS1', 'focused': False, 'urgent': False, 'visible': False},
 			{'name': '2: w2', 'output': 'LVDS1', 'focused': False, 'urgent': False, 'visible': True},
 			{'name': '3: w3', 'output': 'HDMI1', 'focused': False, 'urgent': True, 'visible': True},
 			{'name': '4: w4', 'output': 'DVI01', 'focused': True, 'urgent': True, 'visible': True},
-		]))):
+		])
+
+	def get_outputs(self, pl):
```
ZyX-I: These both are `@staticmethod`.

tests/test_segments.py
S0lll0s: @ZyX-I so for the record, `workspaces` should support both the old and the new functionality for now and be tested against both, `workspace` should only support the new one?

ZyX-I: @S0lll0s `worspaces` has to keep backward compatibility, so both old and new functionality. `workspace` does not have to do this.

ZyX-I: Though since `workspaces` was deprecated I expected it to have *only* old functionality.

S0lll0s:
@ZyX-I not getting notifications anymore?

ZyX-I:
@S0lll0s I never saw notifications regarding new commits. Only messages (comments) and mentions. (Though I am only checking email notifications.)

powerline/segments/i3wm.py
ZyX-I: By the way, what was this `min()` call for originally? `''[10:] == ''`, even though `len('') == 0`.

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,49 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+
+	:param int strip:
+		Specifies how many characters from the front of each workspace name 
+		should be stripped (e.g. to remove workspace numbers). Defaults to zero.
+
+	Highlight groups used: ``workspace`` or ``w_visible``, ``workspace`` or ``w_focused``, ``workspace`` or ``w_urgent``.
+	'''
+	workspace = workspace or segment_info.get('workspace')['name']
+	if segment_info.get('workspace') and segment_info.get('workspace')['name'] == workspace:
+		w = segment_info.get('workspace')
```
ZyX-I: If `segment_info.get('workspace')` is true, then key is present and `.get` is not needed in this line. Also remove `.get()` from the above condition, after `and`.

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,49 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+		Specifies which workspace to show. If unspecified, may be set by the 
+		``list_workspaces`` lister.
+
+	:param int strip:
+		Specifies how many characters from the front of each workspace name 
+		should be stripped (e.g. to remove workspace numbers). Defaults to zero.
+
+	Highlight groups used: ``workspace`` or ``w_visible``, ``workspace`` or ``w_focused``, ``workspace`` or ``w_urgent``.
+	'''
+	workspace = workspace or segment_info.get('workspace')['name']
```
ZyX-I: If `segment_info.get('workspace')` returns None this will throw `TypeError` *before* the next line was checked. Note that code normally does not expect key be absent in `segment_info`. Also if such condition was expected it should have been tested and error found in tests.

ZyX-I: Returns None and `workspace` argument is missing.

ZyX-I: Am I right that there are two valid ways to get workspace: specify it via argument and specify it in segment_info, but segment is not supposed to work when nothing is specified? I would expect it to show current workspace in this case (first current workspace in case there are many due to e.g. multiple monitors).

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,49 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
+	if segment_info.get('workspace') and segment_info.get('workspace')['name'] == workspace:
+		w = segment_info.get('workspace')
+	else:
+		try:
+			w = next((
+				w for w in get_i3_connection().get_workspaces()
+				if w['name'] == workspace
+			))
+		except StopIteration:
+			return
```
ZyX-I: Explicit `None`.

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,49 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 
 @requires_segment_info
+def workspace(pl, segment_info, workspace=None, strip=0):
+	'''Return state for the specified workspace
+
+	:param str workspace:
+		Specifies which workspace to show. If unspecified, may be set by the 
+		``list_workspaces`` lister.
+
+	:param int strip:
```
ZyX-I: `bool`

tests/test_segments.py
```diff
@@ -14,6 +14,7 @@
 from powerline.lib.unicode import out_u
 
 import tests.vim as vim_module
+import powerline.listers.i3wm as i3wm_listers
```
ZyX-I: Unused. And where is the file with the test?

DanielGGordon:
Can uh, someone explain what i3wm workspace thing is?

powerline/segments/i3wm.py
ZyX-I: @S0lll0s That it is not specified in configuration.

S0lll0s:
@DanielGGordon what i3wm workspace thing do you mean? Before this PR there was the [`segments.i3wm.workspaces` segment](http://powerline.readthedocs.org/en/master/configuration/segments/i3wm.html), which shows a list of workspaces of the [i3wm window manager](http://i3wm.org).

This PR introduces a [powerline *lister*](http://powerline.readthedocs.org/en/master/configuration/listers.html) which lists all workspaces, allowing for more detailed custom configurations.

S0lll0s:
@ZyX-I think I got everything once again

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,58 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 @requires_segment_info
+def workspace(pl, segment_info, workspace=None, strip=0):
+	'''Return state for the specified workspace
+
+	:param str workspace:
+		Specifies which workspace to show. If unspecified, may be set by the 
+		``list_workspaces`` lister if used, otherwise falls back to 
+		currently focused workspace.
+
+	:param int strip:
```
ZyX-I: This documentation is a copy of workspases, but behaviour is different.

powerline/segments/i3wm.py
ZyX-I: And default for `strip` should be True or False here.

powerline/segments/i3wm.py
ZyX-I: You also forgot this.

ZyX-I:
I do not see any other problems.

powerline/segments/i3wm.py
S0lll0s: actually they differ:

    def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
        	'''Return list of used workspaces

vs

    def workspace(pl, segment_info, workspace=None, strip=False):
        	'''Return state for the specified workspace

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,58 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 	return [{
-		'contents': w['name'][min(len(w['name']), strip):],
+		'contents': w['name'][strip:],
 		'highlight_groups': calcgrp(w)
-	} for w in conn.get_workspaces()
-		if (not only_show or any(w[typ] for typ in only_show))
-		and (not output or w['output'] == output)
+	} for w in get_i3_connection().get_workspaces()
+		if ((not only_show or any(w[typ] for typ in only_show))
+		and (not output or w['output'] == output))
```
ZyX-I: Now it has no alignment. Can’t you just copy this from the lister?

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,58 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 	]
 
 
 @requires_segment_info
+def workspace(pl, segment_info, workspace=None, strip=False):
+	'''Return state for the specified workspace
```
ZyX-I: > actually they differ:
> 
>     def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
>         	'''Return list of used workspaces
> 
> vs
> 
>     def workspace(pl, segment_info, workspace=None, strip=False):
>         	'''Return state for the specified workspace

I commented near `:param bool strip:` because it is what had the problem. But now that you highlighted it I see that documentation is not valid: `workspace` segment does not return *state* it returns workspace *name*. Usage of different highlight group is irrelevant and should not go into the summary: see how this is described in `powerline.segments.common.env.user`.

S0lll0s:
@ZyX-I sorry about that. Btw what is the preferred format for one-sentence docstrings? I've found examples for both colon-terminated and "free" ones throughout the powerline source.

ZyX-I:
@S0lll0s Colon-terminated? I usually write summaries (not one-sentence docstrings: parameter descriptions have dot even if they contain only one sentence) without dot at the end. At least, I do like this now. Though I would say that in the online documentation with the currently used theme dots look better.

S0lll0s:
@ZyX-I Oh yeah, I meant period/dot

S0lll0s:
Anything else?

powerline/segments/i3wm.py
```diff
@@ -39,27 +48,58 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 	return [{
-		'contents': w['name'][min(len(w['name']), strip):],
+		'contents': w['name'][strip:],
 		'highlight_groups': calcgrp(w)
-	} for w in conn.get_workspaces()
-		if (not only_show or any(w[typ] for typ in only_show))
-		and (not output or w['output'] == output)
+	} for w in get_i3_connection().get_workspaces()
+		if ((not only_show or any(w[typ] for typ in only_show))
+		    and (not output or w['output'] == output))
```
ZyX-I: This finally has correct alignment, but apparently this part was not valid from the very beginning: `} for` has the same indentation as `]`, but `{…}` is *deeper* and thus cannot have the same indentation. This should be

```Python
return [
    {
        'contents': w['name'][strip:],
        'highlight_groups': calcgrp(w)
    }
    for w in get_i3_connection().get_workspaces()
    if (
        (not only_show or any((w[typ] for typ in only_show)))
        and (not output or w['output'] == output)
    )
]
```

Replace four spaces with tabs and do not mess with alignment at all.

Though no, I will change this myself just in case.

You can officially call this doomed list comprehension. clint.py from Neovim with my recently merged modifications would have caught this, but it is for C.

---

BTW, I was enclosing generator in another level of parenthesis always since the very beginning, because something was failing with syntax from your code in Python-2.6. But Python documentation shows that this syntax is OK from [Python-2.4](https://docs.python.org/2.6/whatsnew/2.4.html) where it was introduced. Do you have any idea what was failing?

tests/test_segments.py
```diff
@@ -943,7 +978,6 @@ def test_mode(self):
 		self.assertEqual(i3wm.mode(pl=pl, segment_info={'mode': 'default'}, names={'default': 'test'}), 'test')
 		self.assertEqual(i3wm.mode(pl=pl, segment_info={'mode': 'test'}, names={'default': 'test', 'test': 't'}), 't')
 
-
```
ZyX-I: This change should not be present.

powerline/segments/i3wm.py
S0lll0s: I was eleven when 2.6 was released and am not well-versed in python compatibility issues so I wouldn't know. And I guess I should start linting on my own....

PR #1521, opened 2016-02-07
janitoring: Removed trailing characters in .rst

mrshu:
* This commit removes all unnecessary trailing whitespace from .rst
  files in the repository.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -2,22 +2,22 @@
 How to contribute
 *****************
 
-So you want to contribute to the Powerline project? Awesome! This document 
-describes the guidelines you should follow when making contributions to the 
+So you want to contribute to the Powerline project? Awesome! This document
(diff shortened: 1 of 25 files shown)


ZyX-I:
They indicate that paragraph continues and are present intentionally. See `w` flag in `:h fo-table` in Vim.

mrshu:
Thanks @ZyX-I, this is most probably the first time I might consider whitespaces useful.

ZyX-I:
@mrshu The reason why I use trailing whitespaces to indicate that paragraph continues is that `formatoptions+=a` (i.e. automatic paragraph reformatting) needs to be fought with in any other case. This reformatting is rather convenient with trailing whitespace, without it is a nightmare that tries to shift indent when it is not needed or join separate paragraphs. With whitespace to stop this one just needs to remove the whitespace.

PR #1522, opened 2016-02-07
fix: Battery segment for multiple batteries

mrshu:
* Make battery segment work for multiple batteries by getting data from
  all of them and then computing the weighted average of their actual
  capacities where the weights are the provided percentages of full
  charge.

  For two batteries, therefore, this would look like the following:

        output = (p_1 * c_1 + p_2 * c_2) / (c_1 + c_2)

  where p_n is the percentage of full charge of the n-th battery and c_n
  is the capacity of the n-th battery at full charge.

* A few trailing spaces were removed in the `bat.py` file.

Signed-off-by: mr.Shu <mr@shu.io>

-------------------------------

Note that this fixes #1288 and implements the percentage computation the author of that issue suggested.

PR #1531, opened 2016-02-20
Multiple Battery Support

### LINKED ISSUE #1288: Battery segment doesn't work with dual-battery laptops.
My laptop is Lenovo Thinkpad X240, and it got two batteries, which is `battery_BAT0` and `battery_BAT1` in UPower.
The battery segment of powerline display only the percentage of `battery_BAT0`, where it should be something like:
(`percentage of BAT0` * `capacity of BAT0` + `percentage of BAT1` * `capacity of BAT1`) / ( `capacity of BAT0` + `capacity of BAT1` )

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8130381-battery-segment-doesn-t-work-with-dual-battery-laptops?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mynameisfiber:
Adding multiple battery support to DBUS+UPOWER and /sys battery interfaces.  Fixes #1288 and hopefully mirrors #1522.

### DIFF
--- a/powerline/segments/common/bat.py
+++ b/powerline/segments/common/bat.py
@@ -30,6 +30,7 @@ def _fetch_battery_info(pl):
 			else:
 				devinterface = 'org.freedesktop.DBus.Properties'
 				devtype_name = interface + '.Device'
+				devices = []
 				for devpath in up.EnumerateDevices(dbus_interface=interface):
 					dev = bus.get_object(interface, devpath)
 					devget = lambda what: dev.Get(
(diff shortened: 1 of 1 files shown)


powerline/segments/common/bat.py
```diff
@@ -46,39 +47,62 @@ def _fetch_battery_info(pl):
...
-					)
+								'State'
+							) != 2
+						return (energy * 100.0 / energy_full), state
+					return _flatten_battery
 				pl.debug('Not using DBUS+UPower as no batteries were found')
 
 	if os.path.isdir('/sys/class/power_supply'):
-		linux_capacity_fmt = '/sys/class/power_supply/{0}/capacity'
+		linux_energy_full_fmt = '/sys/class/power_supply/{0}/enery_full'
```
ZyX-I: @mynameisfiber Should not this be `energy`?

powerline/segments/common/bat.py
```diff
@@ -30,6 +30,7 @@ def _fetch_battery_info(pl):
 			else:
 				devinterface = 'org.freedesktop.DBus.Properties'
 				devtype_name = interface + '.Device'
+				devices = []
```
ZyX-I: Have you tested your code with `import dbus` replaced with `raise ImportError`?

powerline/segments/common/bat.py
```diff
@@ -30,6 +30,7 @@ def _fetch_battery_info(pl):
 			else:
```
ZyX-I: Have you tested your code with `import dbus` replaced with `raise ImportError`?

mynameisfiber: I just explicitly passed to the `/sys` method by settings `devices = None`

powerline/segments/common/bat.py
mynameisfiber: No... we need to know the full capacity and the current capacity of each battery to properly weigh them in the average.

ZyX-I: @mynameisfiber This string contains substring `enery`. *Not* `energy`. This is what I mean.

mynameisfiber: Ah! You are completely right... I fixed the filenames in the most recent commit.

ZyX-I: PR was already merged. You need to run new PR.

ZyX-I: (This is also why I asked about testing: it should have failed with incorrect name.)

mynameisfiber: For sure... I ran the tests for this method before implementing the dbus method... must have been a stray replace-all? who knows... it works now.  New PR in order!

PR #1532, opened 2016-02-21
Fixes bug in multi-battery support

mynameisfiber:
Fixes a bug introduced by #1531 where a typo in the filepath for battery energy levels leads to a `FileNotFoundError`.

### DIFF
--- a/powerline/segments/common/bat.py
+++ b/powerline/segments/common/bat.py
@@ -77,8 +77,8 @@ def _fetch_battery_info(pl):
 				pl.debug('Not using DBUS+UPower as no batteries were found')
 
 	if os.path.isdir('/sys/class/power_supply'):
-		linux_energy_full_fmt = '/sys/class/power_supply/{0}/enery_full'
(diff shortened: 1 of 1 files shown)


powerline/segments/common/bat.py
```diff
@@ -49,7 +49,7 @@ def _fetch_battery_info(pl):
 						continue
 					devices.append(devpath)
 					pl.debug('Using DBUS+UPower with {0}', devpath)
-				if devices:
+				if False and devices:
```
ZyX-I: This definitely should not be here.

mynameisfiber: And this is why we do code review.  This was how I triggered the testing of the other segment.  I think now is a good time to say I'm home sick and on a lot of benadryl so a bit of extra code scrutiny is probably in order.

ZyX-I:
You need to squash.

mynameisfiber:
squashed like a bug. :bug:

Issue #1520, opened 2016-01-29
Rename some highlight groups


ZyX-I:
Currently some group names are inconsistent. Specifically `w_…` groups should be `workspace:…`. `superuser` should be `user:root`.

PR #1542, opened 2016-03-09
Add segments.i3wm.scratchpad

s-ol:
Tests are missing but this needs to wait until acrisci/i3ipc-python#26 gets merged anyway

### DIFF
--- a/powerline/config_files/themes/powerline_unicode7.json
+++ b/powerline/config_files/themes/powerline_unicode7.json
@@ -143,6 +143,15 @@
 			"args": {
 				"text": "🖫⃥"
 			}
+		},
+
+		"powerline.segments.i3wm.scratchpad": {
+			"args": {
+				"icons": {
+					"fresh": "●",
+					"changed": "○"
+				}
+			}
 		}
 	}
 }
--- a/powerline/segments/i3wm.py
+++ b/powerline/segments/i3wm.py
@@ -10,7 +10,7 @@ from powerline.bindings.wm import get_i3_connection
 WORKSPACE_REGEX = re.compile(r'^[0-9]+: ?')
 
 
-def calcgrp(w):
+def workspace_groups(w):
 	group = []
 	if w['focused']:
 		group.append('w_focused')
@@ -53,7 +53,7 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
 	return [
 		{
 			'contents': w['name'][strip:],
-			'highlight_groups': calcgrp(w)
+			'highlight_groups': workspace_groups(w)
 		}
 		for w in get_i3_connection().get_workspaces()
 		if ((not only_show or any(w[typ] for typ in only_show))
@@ -97,7 +97,7 @@ def workspace(pl, segment_info, workspace=None, strip=False):
 
 	return [{
 		'contents': format_name(w['name'], strip=strip),
-		'highlight_groups': calcgrp(w)
+		'highlight_groups': workspace_groups(w)
 	}]
 
 
@@ -115,3 +115,41 @@ def mode(pl, segment_info, names={'default': None}):
 	if mode in names:
 		return names[mode]
 	return mode
+
+
+def scratchpad_groups(w):
+	group = []
+	if w.urgent:
+		group.append('scratchpad:urgent')
+	if w.nodes[0].focused:
+		group.append('scratchpad:focused')
+	if w.workspace().name != '__i3_scratch':
+		group.append('scratchpad:visible')
+	group.append('scratchpad')
+	return group
+
+
+SCRATCHPAD_ICONS = {
+	'fresh': 'O',
+	'changed': 'X',
+}
+
+
+def scratchpad(pl, icons=SCRATCHPAD_ICONS):
+	'''Returns the windows currently on the scratchpad
+
+	:param dict icons:
+		Specifies the strings to show for the different scratchpad window states. Must 
+		contain the keys ``fresh`` and ``changed``.
+
+	Highlight groups used: ``scratchpad`` or ``scratchpad:visible``, ``scratchpad`` or ``scratchpad:focused``, ``scratchpad`` or ``scratchpad:urgent``.
+	'''
+
+	return [
+		{
+			'contents': icons.get(w.scratchpad_state, icons['changed']),
+			'highlight_groups': scratchpad_groups(w)
+		}
+		for w in get_i3_connection().get_tree().descendents()
+		if w.scratchpad_state != 'none'
+	]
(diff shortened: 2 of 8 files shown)


powerline/segments/i3wm.py
```diff
@@ -115,3 +115,37 @@ def mode(pl, segment_info, names={'default': None}):
...
+
+
+def scratchpad_groups(w):
+	group = []
+	if w.urgent:
+		group.append('s_urgent')
+	if w.nodes[0].focused:
+		group.append('s_focused')
+	if w.workspace().name != '__i3_scratch':
+		group.append('s_visible')
```
ZyX-I: Group must be named `scratchpad:visible`, same for other. `w_*` are also invalid names, but they cannot be changed for backwards compatibility #1520.

powerline/segments/i3wm.py
```diff
@@ -115,3 +115,37 @@ def mode(pl, segment_info, names={'default': None}):
...
+	group.append('scratchpad')
+	return group
+
+
+def scratchpad(pl, fresh_icon=u'\u25cf', changed_icon=u'\u25cb'):
```
ZyX-I: Default icons must be ASCII. Unicode strings must not have `u` prefix, there is `from __future__ import unicode_strings`. They must also be placed into all theme files, even if theme defines them identical to default ones.

powerline/segments/i3wm.py
s-ol: ah yes, I forgot the reason they were still called that

powerline/segments/i3wm.py
ZyX-I: Ah, and never use `\u…` notation for strings, if you search the whole repository for `\uXXXX` (e.g. `ag '\\u[0-9a-fA-F]{4}'` you will see it only in regexes and tests). This is not readable and the whole point of powerline is to present something to user, thus icons that are more readable in `\uXXXX` notation may not be accepted.

powerline/config_files/themes/powerline_unicode7.json
```diff
@@ -143,6 +143,15 @@
 			"args": {
 				"text": "🖫⃥"
 			}
+		},
+
+		"powerline.segments.i3wm.scratchpad": {
+			"args": {
+				"icons": {
+					"fresh": "O",
+					"changed": "X"
```
s-ol: apparently `vim` doesn't support utf-7 so I put ASCII symbols instead

ZyX-I: How is UTF-7 related? To get a unicode character given its code there is `<C-v>uXXXX`.

ZyX-I: And `<C-r>=nr2char(0xXXXX)<CR>`. I usually use the latter because it allows correcting errors.

ZyX-I: `unicode7` has nothing to do with UTF-7, it is a file which may contain code points from unicode-7.0 standard.

Icons here *may* be the new ones introduced by the standard, not *must be*. If you do not know newer code points which are more relevant, then you should have characters from powerline theme here.

s-ol: wait, is this file for `UTF-7` or for `Unicode 7.0` symbols? Either way almost none of the other symbols are displayed correctly in vim (or in my browser) so I assumed that if I pasted mine in and they showed up that would mean they are wrongly encoded.

powerline/segments/i3wm.py
```diff
@@ -10,10 +10,9 @@
 WORKSPACE_REGEX = re.compile(r'^[0-9]+: ?')
 
 
-def calcgrp(w):
+def workspace_groups(w):
 	group = []
-	if w['focused']:
-		group.append('w_focused')
+	if w['focused']: group.append('w_focused')
```
ZyX-I: Revert, this should not be one line.

s-ol:
travis fail seems to be random fail? it can't `import unittest2`

powerline/config_files/themes/powerline_unicode7.json
ZyX-I: Unicode 7.0. UTF-7 is *ASCII only* encoding scheme, it *cannot* produce symbols that Vim cannot render unless Vim parsed file as UTF-7. And many symbols are common with powerline theme.

ZyX-I: Purpose of `ascii.json` theme is to use only characters which are valid in ASCII for use in environments where locale is not UTF-8. Switching to ascii theme *will not* cause powerline output only ASCII characters, only locale or relevant option (e.g. `&encoding` in Vim) can do such a thing. Just when you switch to ascii theme you will not get question marks for branch name prefixes because encoding unicode string to `latin1` (should locale be latin1) is a destructive operation.

powerline/segments/i3wm.py
s-ol: accidental change

powerline/segments/i3wm.py
```diff
@@ -115,3 +115,41 @@ def mode(pl, segment_info, names={'default': None}):
...
+SCRATCHPAD_ICONS = {
+	'fresh': 'O',
+	'changed': 'X',
+}
+
+
+def scratchpad(pl, icons=SCRATCHPAD_ICONS):
+	'''Returns the windows currently on the scratchpad
+
+	:param string icons:
```
ZyX-I: string?

s-ol: whoops

s-ol:
uhm.... no idea why vim tests are failing now

ZyX-I:
This may have something to do with me restarting powerline/bot-ci build.

ZyX-I:
Looks like it is because Command-T changed used file name. Will fix later.

s-ol:
the i3ipc-python changes are merged and released to i3ipc, so technically ready to merge.

tests/test_segments.py
```diff
@@ -977,6 +969,45 @@ def test_mode(self):
...
+				workspace_scratch = lambda: Args(name='__i3_scratch')
+				workspace_noscratch = lambda: Args(name='2: www')
+				return [
+					Args(scratchpad_state='fresh', urgent=False, workspace=workspace_scratch, nodes=nodes_unfocused),
+					Args(scratchpad_state='changed', urgent=True, workspace=workspace_noscratch, nodes=nodes_focused),
+					Args(scratchpad_state='fresh', urgent=False, workspace=workspace_scratch, nodes=nodes_unfocused),
+					Args(scratchpad_state=None, urgent=False, workspace=workspace_noscratch, nodes=nodes_unfocused),
+					Args(scratchpad_state='fresh', urgent=False, workspace=workspace_scratch, nodes=nodes_focused),
+					Args(scratchpad_state=None, urgent=True, workspace=workspace_noscratch, nodes=nodes_unfocused),
+				]
```
ZyX-I: Missing empty line. AFAIR linter does require it here.

s-ol: there was one line too much below too, fixed both

PR #1548, opened 2016-03-19
fix asrun to pass bytes instead of a string to run_cmd

anilanar:


powerline/lib/shell.py
```diff
@@ -43,7 +43,7 @@ def run_cmd(pl, cmd, stdin=None, strip=True):
 
 def asrun(pl, ascript):
 	'''Run the given AppleScript and return the standard output and error.'''
-	return run_cmd(pl, ['osascript', '-'], ascript)
+	return run_cmd(pl, ['osascript', '-'], ascript.encode('utf-8'))
```
ZyX-I: You should use `get_preferred_output_encoding` from `powerline.lib.encoding`, not `utf-8`. Also if `Popen.communicate` is not accepting bytes, this should be fixed in `run_cmd`. At the current state it is safe to assume that `run_cmd` receives only unicode or `None` arguments in `stdin`.

PR #1549, opened 2016-03-19
Make Popen.communicate receive only bytes

### LINKED ISSUE #1547: run_cmd: Popen.communicate doesn't take a string
`powerline.lib.shell.run_cmd` function calls `Popen.communicate(stdin)`.
`powerline.lib.shell.asrun` function calls `run_cmd` and passes `ascript` to its 3rd argument (`stdin`).
`powerline.segments.common.spotify` calls `asrun` with an apple script of type "string".

But according to Python [2.7](https://docs.python.org/2/library/subprocess.html) and [3.5](https://docs.python.org/3/library/subprocess.html) docs, `Popen.communicate` takes a "data", not a "string".

So as a result, `spotify_apple_script` doesn't work and complain about expecting bytes but receiving str, throws an error.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32007082-run_cmd-popen-communicate-doesn-t-take-a-string?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F283196&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ZyX-I:
Fixes #1547
Closes #1548

### DIFF
--- a/powerline/lib/shell.py
+++ b/powerline/lib/shell.py
@@ -7,7 +7,7 @@ import os
 from subprocess import Popen, PIPE
 from functools import partial
 
-from powerline.lib.encoding import get_preferred_input_encoding
+from powerline.lib.encoding import get_preferred_input_encoding, get_preferred_output_encoding
 
 
 if sys.platform.startswith('win32'):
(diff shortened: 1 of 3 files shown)


anilanar:
Solves the problem, thanks!

PR #1551, opened 2016-03-22
More specific instructions for iPython

jtyr:
This PR is improving documentation about how to integrate Powerline with iPython. Tested on Arch Linux and Ubuntu.


ZyX-I:
PRs to any branch, but develop, are not accepted.

ZyX-I:
Note that `ipython_config.py` file is not necessary located in `~/.ipython/profile_default`. There may be more then one profile.

jtyr:
People are using the default profile in 99% of cases. Who is not using it, then probably knows that it must be changed in the path.

I have created PR #1552 against the `develop` branch.

PR #1557, opened 2016-03-24
Fix OS X installation guide little formatting issues and phrasing

gianpaj:


docs/source/installation/osx.rst
```diff
@@ -61,9 +59,9 @@ but MacVim users need to install it using the following command::
 Fonts installation
 ==================
 
-Install downloaded patched font by double-clicking the font file in Finder, then 
-clicking :guilabel:`Install this font` in the preview window.
+Download the patched font from Github, install it by double-clicking the font file in Finder and then
```
ZyX-I: This line is too long.

docs/source/installation/osx.rst
```diff
@@ -24,28 +22,28 @@ Python package
 
 2. Install Powerline using one of the following commands:
 
+    This will get the current release version:
```
ZyX-I: This line looks clumsy with the preceding line. Also indentation is wrong: should be three spaces. Previous variant that formed a sentence with code blocks looks better.

docs/source/installation/osx.rst
```diff
@@ -24,28 +22,28 @@ Python package
...
 
    .. code-block:: sh
 
        pip install --user git+git://github.com/powerline/powerline
 
-   will get latest development version.
-
    .. warning::
-      When using ``brew install`` to install Python one must not supply
+      When using ``brew install`` to install Python, one must not supply
```
ZyX-I: I am very unsure that comma is needed here and in the next clause.

docs/source/installation/osx.rst
ZyX-I: And do not mention github here, it is not the *only* possible source of patched font downloads and it is absolutely irrelevant from where patched font was downloaded.

ZyX-I: It is actually not necessary that font was downloaded at all. Existing phrase is good except that “downloaded” does not have to be there.

ZyX-I:
Actually there are only three changes which may be accepted without questions: removal of two `.` and space before <code>``</code>. I will apply them manually.

PR #1561, opened 2016-03-29
Increase socket backlog number for `powerline-daemon`

iblis17:
The number `1` for backlog is quite small.

On my FreeBSD box, I got this message:
```
Mar 28 19:01:43 imfsa kernel: sonewconn: pcb 0xfffff802088b25a0: Listen queue overflow: 2 already in queue awaiting acceptance (32 occurrences)
```
and
```
% netstat -na | grep fffff802088b25a0
ffffff802088b25a0 stream      0      0 fffff802088b25a0        0        0        0 /tmp/powerline-ipc-1001
```

Increasing the backlog number can solve the queue overflow problem.

### DIFF
--- a/scripts/powerline-daemon
+++ b/scripts/powerline-daemon
@@ -231,7 +231,7 @@ def do_one(sock, read_sockets, write_sockets, result_map):
 
 
 def main_loop(sock):
-	sock.listen(1)
+	sock.listen(128)
 	sock.setblocking(0)
 
 	read_sockets, write_sockets = set(), set()


ZyX-I:
According to the Python-2.6 `sock.listen` documentation

> Listen for connections made to the socket. The backlog argument specifies the maximum number of queued connections and should be at least 1; the maximum value is system-dependent (usually 5).

. Will not this change break some systems? “The maximum value” here looks like it will throw if it is exceeded, so you should probably use `5`. Or, if you know that FreeBSD always allows 128, limit this change to FreeBSD only.

ZyX-I:
(Interesting that Python-3\* documentation does not talk about any upper limits.)

iblis17:
I checked FreeBSD 9.3, 10.1, 11(current).
There is a kernel option on 9, `kern.ipc.somaxconn`. The default is 128.
After 10(including 10), `somaxconn` is renamed to `soacceptqueue`. The default is also 128.
FYI: https://www.freebsd.org/cgi/man.cgi?query=listen&sektion=2#end

About linux... `cat /proc/sys/net/core/somaxconn` -> 128 by default.

I do no check windows yet(There isn't any win in my env).
FYI:http://stackoverflow.com/questions/4709756/listen-maximum-queue-size-per-windows-version
(I cannot figure out the correct number :sob:

I think we have two solutions to deal with the backlog.
1. Just bump it to 128 first, put it in error handling flow. If listen failed, then change to 5.
2. Make this number configurable. Get it via env var or powerline `config.json`.
Which one is better?

ZyX-I:
@iblis17 Actually on my system it accepts even `2**31 - 1` (i.e. INT_MAX). I have no idea whether this will (not) fail on other systems, 90% that number used as input is simply passed to the underlying C function, `man 2 listen` explicitly specifies that it silently uses `min(backlock, /proc/sys/net/core/somaxconn)`. If BSD man page you linked had the same statement I think it probably makes sense to simply accept this change and check whether anybody complains. I do not see it, so need to ask to test: check with `2**31-1` as well.

Daemon is not used on Windows (no applications for which it may be used) and most likely does not work, so you do not need to care.

iblis17:
I got the source code:
https://svnweb.freebsd.org/base/releng/10.3/sys/kern/uipc_socket.c?revision=296373&view=markup#l674

I also tested the `INT_MAX` on freebsd, same behavior as linux!

yonas:
@iblis17 Thanks! This error was filling up my logs.

ZyX-I:
You need to rebase: at least some tests are failing because of the unrelated error.

iblis17:
rebased !

PR #1563, opened 2016-04-03
Official name for "command" mode is "normal" mode -> unified descript…

GHPS:
…ion that powerline also uses in vim.


ZyX-I:
`man zshzle`, section KEYMAPS:

> **vicmd**  vi emulation  - command mode

. I am *not* using *Vim* terminology. I am using whatever terminology is used in the application bindings are created for.

PR #1565, opened 2016-04-06
Fix and update documentation concerning i3wm

s-ol:
update dependencies for xrandr and i3ipc, fix a typo

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -30,8 +30,9 @@ Generic requirements
   with bazaar repositories.
 * ``pyuv`` python package. Required for :ref:`libuv-based watcher 
   <config-common-watcher>` to work.
-* ``i3-py``, `available on github <https://github.com/ziberna/i3-py>`_. Required 
-  for i3wm bindings and segments.
+* ``i3-ipc`` python package. Required for i3wm bindings and segments.
+* ``xrandr`` program. Required for the multi-monitor lemonbar binding and the 
+  :py:func:`powerline.listers.i3wm.output_lister`.
 
 .. note::
     Until mercurial and bazaar support Python-3 or PyPy powerline will not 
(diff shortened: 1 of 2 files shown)


docs/source/installation.rst
```diff
@@ -30,8 +30,9 @@ Generic requirements
   with bazaar repositories.
 * ``pyuv`` python package. Required for :ref:`libuv-based watcher 
   <config-common-watcher>` to work.
-* ``i3-py``, `available on github <https://github.com/ziberna/i3-py>`_. Required 
-  for i3wm bindings and segments.
+* ``i3-ipc`` python package. Required for i3wm bindings and segments.
+* ``xrandr`` program. Required for the multi-monitor lemonbar binding and the 
+  :func:`i3wm output_lister <powerline.listers.i3wm.output_lister>`.
```
ZyX-I: Are you sure this reference works? I only ever used <code>:py:func:\`powerline.listers.i3wm.output_lister\`</code> (note: `:py:`, also no `<>`).

s-ol: I tried it locally and it did, but I can match that style.

It looked a bit weird though, "output_lister" is written in monospaced font with a well around it, it doesn't look very "clickable". Is :py:func better?

ZyX-I: I used `:py:func:` simply because after reading sphinx doc I got the impression that `:py:func:` is exactly what I should use.

(“Did” means that reference pointed to a right location?)

s-ol:
I suppose travis is still failing because of the Ctrl-P (or whatever it was) bug?

ZyX-I:
Travis is still failing because I do not have #1553 finished (in fact I have no idea why fix there is not enough).

s-ol:
I changed it to `:py:func:` for consistency, it looks nearly the same (only adds `()` behind it).

PR #1567, opened 2016-04-08
Add `short` param for system_load segment

iblis17:


ZyX-I:
Missing tests.

iblis17:
@ZyX-I ok. I added the tests!

powerline/segments/common/sys.py
```diff
@@ -13,7 +13,8 @@
 cpu_count = None
 
 
-def system_load(pl, format='{avg:.1f}', threshold_good=1, threshold_bad=2, track_cpu_count=False):
+def system_load(pl, format='{avg:.1f}', threshold_good=1, threshold_bad=2,
+		track_cpu_count=False, short=False):
```
ZyX-I: You should either leave everything on one line or align with first argument (note: *alignment* uses spaces).

iblis17: hmm... I just found that this code is using tab for indentation.
So, will using all spaces for this line be suitable? Or to keep this consistency in same file, should i leave it without new line?

iblis17:
@ZyX-I ok. I got it.
I just figured it out from the `CONTRIBUTING.rst`.

Issue #1571, opened 2016-04-16
Weather segment no longer works due to Yahoo requiring OAuth


DomBlack:
I just tried adding the weather segment, and got this error. After looking at it, it turns out that Yahoo has forced OAUTH on their weather API's.

```
2016-04-16 09:41:33,504:ERROR:tmux:WeatherSegment:Yahoo returned malformed or unexpected response: u'{"query":{"count":1,"created":"2016-04-16T08:41:33Z","lang":"en-US","results":{"weather":{"xmlns":"http://where.yahooapis.com/v1/schema.rng","error":{"lang":"en-US","yahoo":"http://www.yahooapis.com/v1/base.rng","uri":"http://yahoo.com","description":"Please provide valid credentials. OAuth oauth_problem=\\"OST_OAUTH_PARAMETER_ABSENT_ERROR\\", realm=\\"yahooapis.com\\"","detail":"Please provide valid credentials. OAuth oauth_problem=\\"OST_OAUTH_PARAMETER_ABSENT_ERROR\\", realm=\\"yahooapis.com\\""}}}}}'
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/powerline/segments/common/wthr.py", line 146, in compute_state
    condition = response['query']['results']['weather']['rss']['channel']['item']['condition']
KeyError: u'rss'
```

gsusrafael:
i've solved this issue right now using geoip.nekudo.com, a custom YQL query and removing the call of weather.bylocation.xml remote table. I'm posting my patch now:

```patch
--- a/wthr.py.orig
+++ b/wthr.py.new
@@ -115,19 +115,20 @@ class WeatherSegment(KwThreadedSegment):
                        return self.location_urls[location_query]
                except KeyError:
                        if location_query is None:
-                               location_data = json.loads(urllib_read('http://freegeoip.net/json/'))
+                               location_data = json.loads(urllib_read('http://geoip.nekudo.com/api/'))
                                location = ','.join((
                                        location_data['city'],
-                                       location_data['region_name'],
-                                       location_data['country_code']
+                                       location_data['country']['name'],
+                                       location_data['country']['code']
                                ))
-                               self.info('Location returned by freegeoip is {0}', location)
+                               #self.info('Location returned by freegeoip is {0}', location)
                        else:
                                location = location_query
+
                        query_data = {
                                'q':
[...]

PR #1596, opened 2016-05-11
segments/weather: improve weather query

ZyX-I:
The weather query no longer depends on an external lookup table for
location codes.


ZyX-I:
Ref #1571.

PR #1604, opened 2016-05-27
Allow for differently coloured email notifications for multiple mail …

fischmax:
…accounts

Right what it says on the box ;-)

### DIFF
--- a/powerline/segments/common/mail.py
+++ b/powerline/segments/common/mail.py
@@ -37,18 +37,18 @@ class EmailIMAPSegment(KwThreadedSegment):
 		return unread_count
 
 	@staticmethod
-	def render_one(unread_count, max_msgs=None, **kwargs):
+	def render_one(unread_count, max_msgs=None, highlight_group='email_alert', highlight_group_gradient='email_alert_gradient', **kwargs):
 		if not unread_count:
 			return None
 		elif type(unread_count) != int or not max_msgs:
 			return [{
 				'contents': str(unread_count),
-				'highlight_groups': ['email_alert'],
+				'highlight_groups': [highlight_group],
 			}]
 		else:
 			return [{
 				'contents': str(unread_count),
-				'highlight_groups': ['email_alert_gradient', 'email_alert'],
+				'highlight_groups': [highlight_group_gradient_, highlight_group],
 				'gradient_level': min(unread_count * 100.0 / max_msgs, 100),
 			}]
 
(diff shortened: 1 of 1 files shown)


powerline/segments/common/mail.py
```diff
@@ -73,6 +73,14 @@ def render_one(unread_count, max_msgs=None, **kwargs):
...
+	Highlight group to use. If not present, ``email_alert`` will be used. This
+	parameter allows you to watch over multiple mailboxes with differently
+	coloured notifications.
+:param str highlight_group_gradient_:
+	Highlight group to use for gradient, i.e. when ``max_msgs`` is given. If
+	not present, ``email_alert_gradient`` will be used.
 
-Highlight groups used: ``email_alert_gradient`` (gradient), ``email_alert``.
+Highlight groups used: ``email_alert_gradient`` (gradient), ``email_alert`` or
+	the given ones.
```
ZyX-I: This line is for the linter. It *must not* be modified like this.

ZyX-I:
I do not like the idea of defining highlight group in the function arguments, it breaks linter and it is inconsistent with other code. I would suggest to simply use different text in `before`/`after` in place of different highlighting, there is not much sense in designing something (e.g. `highlight_overrides` in configuration which will work for any segment) just for such rare case.

fischmax:
> there is not much sense in designing something [...] just for such rare case.

Mkay, I'll just use my fork then for my „rare“ case ;-)

BlitzKraft:
@ZyX-I 
This is not a rare case. I was looking for something like this when I first started using powerline. I settled for a solution - one account on the leftside and another on the right side of tmux status. However, I would have preferred to be able to add a prefix/suffix, or use a different color. 

On a more general case, for having multiple instances of same segments, there is not an easy way to distinguish one from another. Example, two music player segments, one with spotify and another with mpd playing locally. 

There needs to be a way to specify highlight groups, just so manual tweaking of colorschemes on a per segment basis would be possible.

ZyX-I:
@BlitzKraft Prefix/suffix on per-segment basis is always possible. Highlighting is not. If somebody wants to implement that suggestion with

```JSON
{"segments":{"left":[{
    …
    "highlight_overrides": {
        "email_alert": "email_alert_2",
        "email_alert_gradient": "email_alert_gradient_2"
    }
}]}}
```

I will merge this, but not PR which adds highlight group to arguments. And I will not write this myself (and note that PR with highlight overrides will need tests (rather trivial) and non-trivial changes to linter). Case is rare because you need from the number of people who need different instances of one segment (I can imagine only email and weather, no idea why multiple player segments are needed) substract the number of people who are fine with marking them with different text and not with different highlighting.

PR #1607, opened 2016-06-04
Improve weather query

pthomas551:
#1596 does not seem to have been merged so opening this PR to remove use of the external lookup table for location_query.


ZyX-I:
That was merged.

PR #1614, opened 2016-07-06
Fix trailing whitespace segment on Python 3.

### LINKED ISSUE #1613: Trailing whitespace segment does not work on Python 3
The trailing whitespace segment indicates the wrong lines, due to [this line](https://github.com/powerline/powerline/blob/develop/powerline/segments/vim/__init__.py#L581). Running `str` on `bytes` does not automatically encode anything anymore. `str(b' \t')` creates a string `"b' \\t'"`, which means it doesn't just match whitespace, but also the letters b and t, plus a single quote and backslash.

QuLogic:
Using str() on a bytestring produces the repr on Python 3, not some
implicitly decoded string. So use an explicit decode instead.

Fixes #1613.

### DIFF
--- a/powerline/segments/vim/__init__.py
+++ b/powerline/segments/vim/__init__.py
@@ -578,7 +578,7 @@ def trailing_whitespace(pl, segment_info):
 	else:
 		buf = segment_info['buffer']
 		bws = b' \t'
-		sws = str(bws)
+		sws = str(' \t')  # Ignore unicode_literals and use native str.
 		for i in range(len(buf)):
 			try:
 				line = buf[i]


ZyX-I:
@QuLogic `buf[i]` produces `str()` both on Python and Python3. This is why I used `str()` and not e.g. `u`, `b'\xAB' in u' '` is UnicodeDecodeError: this is exactly what you will get should you encounter the line that ends with `«` (as an example, there will be problems with any multibyte character).

QuLogic:
> buf[i] produces str() both on Python and Python3.

OK, then I don't really understand why there's a `try: use str except UnicodeDecodeError: use bytes` if it's always a `str`, but that doesn't really affect the patch

> This is why I used str() and not e.g. u, b'\xAB' in u' ' is UnicodeDecodeError: this is exactly what you will get should you encounter the line that ends with « (as an example, there will be problems with any multibyte character).

Yes, of course `b'\xAB' in u' '` would not work, but you said `buf[i]` is a `str`, so it'd be `u' ' in u' '`, which would work fine.

ZyX-I:
@QuLogic No, it would not. Python 2’s `str()` is `bytes()`, not `unicode()`. Python 3’s `str()` is (removed) `unicode()`, but here `buf[i]` may produce UnicodeDecodeError in case buffer contains invalid unicode characters.

ZyX-I:
Most logical here is not bother with encoding/decoding and just write `str(' \t')` (without `b` string prefix). Or `str(bws.encode())`.

ZyX-I:
I.e. `buf[i]` returns `str()` on both Python versions, but `str(' \t')` is *not* always `u' \t'`, in Python 2 it is `b' \t'`.

QuLogic:
Oh, I misunderstood under which version you meant the `UnicodeDecodeError` would occur. The cast should work in that case.

PR #1618, opened 2016-07-08
Add support for IPython-5

### LINKED ISSUE #1617: Powerline doesn't work with IPython 5.x
Just after run ipython one can see this:

```python
[TerminalIPythonApp] WARNING | Error in loading extension: powerline.bindings.ipython.post_0_11
Check your config files in /home/local/GER/psocha/.ipython/profile_default
  File "/usr/local/lib/python2.7/dist-packages/IPython/core/shellapp.py", line 246, in init_extensions
    self.shell.extension_manager.load_extension(ext)
  File "/usr/local/lib/python2.7/dist-packages/IPython/core/extensions.py", line 83, in load_extension
    __import__(module_str)
  File "/usr/local/lib/python2.7/dist-packages/powerline/bindings/ipython/post_0_11.py", line 6, in <module>
    from IPython.core.prompts import PromptManager
ImportError: cannot import name PromptManager
```

Quick grep through IPython sources shows:

```python
        if 'PromptManager' in self.config:
            warn('As of IPython 5.0 `PromptManager` config will have no effect'
                 ' and has been replaced by TerminalInteractiveShell.prompts_class')
```

ZyX-I:
Fixes #1617

### DIFF
--- a/powerline/bindings/ipython/post_0_11.py
+++ b/powerline/bindings/ipython/post_0_11.py
@@ -2,12 +2,12 @@
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
 from weakref import ref
+from warnings import warn
 
 try:
 	from IPython.core.prompts import PromptManager
 	has_prompt_manager = True
 except ImportError:
-	from IPython.terminal.interactiveshell import TerminalInteractiveShell
 	has_prompt_manager = False
 from IPython.core.magic import Magics, magics_class, line_magic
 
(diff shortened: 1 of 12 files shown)


ZyX-I:
Collection of fixed problems:

1. Highlighting is using pygments tokens which (much like Vim highlight groups) are to define *what* is highlighted, while powerline defines *how*. Solution: generate names and use custom style.
2. Custom style cannot be registered and used as a regular pygments style because powerline defines only style that can return only highlighting for powerline-specific tokens. Solution: use hacks to alter `TerminalInteractiveShell._style` attribute: one to make changes to style work (that is override of `_make_style_from_name` function) and another to make it work at startup (that is immediate `ip._style` setting in `ConfigurableIpythonPowerline.do_setup`).
3. Powerline renderer was always supposed to output only strings. Solution: that was not much of a problem though, replacing a few occurrences of `''.join` in `powerline.renderer` module did the trick.


Not as much as I thought, most time wasted to figure out what I need to change. At least I now do not need to push unicode through code that actively wants ASCII only even if it is actually capable of dealing with unicode.

PR #1621, opened 2016-07-18
reset last status code before calling zle reset-prompt

AprilArcus:
Currently, toggling modes in the Zsh line editor causes the `powerline.segments.shell.last_status` segment to be incorrectly cleared, because the return code of `_powerline_set_true_keymap_name $KEYMAP` fills the value of `$?` at draw time. This patch to `bindings/zsh/powerline.zsh` captures the last status code and restores it before redrawing the prompt.

### DIFF
--- a/powerline/bindings/zsh/powerline.zsh
+++ b/powerline/bindings/zsh/powerline.zsh
@@ -74,7 +74,9 @@ _powerline_init_modes_support() {
 	}
 
 	function -g _powerline_zle_keymap_select() {
+		local last_status=$?
 		_powerline_set_true_keymap_name $KEYMAP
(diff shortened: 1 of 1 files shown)


ZyX-I:
I am unable to reproduce, `last_status` correctly shows last status when changing modes, as well as `last_pipe_status`. Zsh version is 5.2.

`zle reset-prompt` should already fill the correct values of `$?` and `$pipestatus` before redrawing prompt.

Also there is no need in any subshells to set `$?`: just `() return last_status` is enough. I am not accepting this patch though, it does not restore `$pipestatus` and zsh authors should have already cared about this.

PR #1641, opened 2016-08-18
Python process replication fix while restarting Awesome 

pozar87:
Hello,

I run into an issue with **powerline-awesome.py** replication while restarting **Awesome Window Manager**. After some investigation I've found that `ps -C powerline-awesome.py` is not returning proper results. At least in my version of `ps` (I'm using **Arch**):

> ps --version 
> ps from procps-ng 3.3.12

So I modified the `powerline/bindings/awesome/powerline.lua` a bit to use `ps aux` and `grep`  (which should work similarly on all Linux distributions - correct me if I'm wrong).

Finally I created this pull request to share my experience - feel free to reject it if it's useless :)

Lukasz

PS Keep up the good work! Powerline is really awesome!

### DIFF
--- a/powerline/bindings/awesome/powerline.lua
+++ b/powerline/bindings/awesome/powerline.lua
@@ -8,4 +8,4 @@ function powerline(mode, widget) end
 
 bindings_path = string.gsub(debug.getinfo(1).source:match('@(.*)$'), '/[^/]+$', '')
 powerline_cmd = bindings_path .. '/powerline-awesome.py'
-awful.util.spawn_with_shell('ps -C powerline-awesome.py || ' .. powerline_cmd)
+awful.util.spawn_with_shell('ps aux | grep -e powerline-awesome.py | grep -v grep || ' .. powerline_cmd)


ZyX-I:
This is not going to work correctly. If you e.g. run `vim path/to/powerline-awesome.py` then `ps aux | grep -e powerline-awesome.py | grep -v grep` will succeed and script will not be run. I guess the problem is that in your case command name is something like `python` (this is not uncommon).

Actually also neither your nor original variants are correct: they disallow two `powerline-awesome.py` processes run by different users. It needs a different fix.

pozar87:
You are right, I missed the obvious... 
I'm going to give this a second thought this weekend. It would be good to solve this once and for all. Should I modify this pull request with better implementation or create a bug ticket and you guys will take care of that?

ZyX-I:
@pozar87 Can you check whether #1642 works? In any case this is roughly how I imagine this (also for other WM widgets).

ZyX-I:
Note: there are no checks regarding whether “daemon is running” or “daemon is running wm bindings” in lua script: daemon cares about both (`powerline-daemon -q` it will not do anything if socket was already taken and `powerline wm.awesome` will not make daemon run bindings twice because of threads caching).

PR #1642, opened 2016-08-19
Use powerline daemon for running awesome bindings

ZyX-I:
Also adds functional tests for awesome bindings.

Replaces #1641

### DIFF
--- a/docs/source/configuration/reference.rst
+++ b/docs/source/configuration/reference.rst
@@ -215,6 +215,13 @@ Common configuration is a subdictionary that is a value of ``ext`` key in
 
     All components are enabled by default.
 
+.. _config-ext-update_interval:
+
+``update_interval``
+    Determines how often WM status bars need to be updated, in seconds. Only 
+    valid for WM extensions which use ``powerline-daemon``. Defaults to 
+    2 seconds.
+
 .. _config-colors:
 
 Color definitions
--- a/powerline/__init__.py
+++ b/powerline/__init__.py
@@ -409,6 +409,11 @@ def _get_log_keys(common_config):
 	))
 
 
+DEFAULT_UPDATE_INTERVAL = 2
+'''Default value for :ref:`update_interval <config-ext-update_interval>`
+'''
+
+
 class Powerline(object):
 	'''Main powerline class, entrance point for all powerline uses. Sets 
 	powerline up and loads the configuration.
@@ -504,6 +509,7 @@ class Powerline(object):
 		self.setup_args = ()
 		self.setup_kwargs = {}
 		self.imported_modules = set()
+		self.update_interval = DEFAULT_UPDATE_INTERVAL
 
 	get_encoding = staticmethod(get_preferred_output_encoding)
 	'''Get encoding used by the current application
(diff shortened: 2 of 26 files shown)


pozar87:
The new way seems to work. Restarting **Awesome** is not triggering additional python processes - Cool!
However I run int a small issue - and I'm not yet sure what is the reason on this. 
In short `/usr/bin/python /usr/bin/powerline-daemon -q` is starting with **Awesome**, but no Powerline widget is added. After I restart **Awesome** I can see some (four to be precised ) sup-process of `powerline-daemon` in `htop` and then widget is there. 
I suspect I need to modify some integration code in **Awesome** lua files?

ZyX-I:
`htop` shows threads separately AFAIR, and powerline-daemon uses some.

Though I guess what is the issue here: need `sleep()` for daemon to have some time to initialize Python and actually open socket `powerline` can connect to. Do you know how to do this?

pozar87:
Yes, I also suspect some race condition here. I'll try to debug how (and when) the **Awesome** is loading the `powerline` widget, perhaps I can add some loop with a few retries there.

pozar87:
Hi, 
I cannot see any obvious way where I can put some retry/fallback code for `powerline`:

```
package.path = package.path .. ';/usr/lib/python3.5/site-packages/powerline/bindings/awesome/?.lua'
require('powerline')
(...)
mywidgets = {
  powerline_widget,
  (...)
}
```
And that's it... Is there any method I can call form lua to check if `powerline` hes been initialized? (this however complicate the required **Awesome** integration)

ZyX-I:
@pozar87 Check whether AF_UNIX socket bound to `powerline-ipc-{user ID}` address can be connected to. Why not just add `sleep()`, awesome library also does not provide this feature?

pozar87:
Hi, 
I was experimenting with something like:
```
for i=1,10 do 
  if os.execute("ss -a | grep powerline && sleep 0.2") then 
    break
  end	
end
```
 As running `sleep()` in **lua** seems to be not straightforward... and it feels better to have some lock base on condition not just time.

But I cannot make it work. I'm quite sure the code is being executed correctly, however widget is not showing untill restart (when` /usr/bin/powerline-daemon -q` is already up and running). Do you think I should contact the **Awesome Window Manager** community? 

PS On the bright site - your solution of the original process duplication problem seems to be solved in quite bullet-proof way :)

ZyX-I:
@pozar87 If you think it is needed, better contact them. I personally am not using `awesome`, so debugging capabilities are rather limited.

How did you make sure that code is being executed correctly? I would try using `strace -f -e execve` to start awesome. Also wondering whether `awful.util.spawn_with_shell` is (not) blocking. Maybe `powerline wm.awesome` should actually never finish, like previous solution did.

Also it seems that using `ss` is not a good idea: 1. I have it in /sbin (so it is supposed to be used by root only) and 2. as usual with “list everything and grep” solution I immediately found socket which has nothing to do with daemon: `/tmp/screen/S-zyx/7787.powerline` (except that it was used to run konsole with tmux client inside with multiple shells run which have working directory in powerline repository).

Issue #1119, opened 2014-10-15
Weather segment sporadic errors.


jottr:
The weather segment sporadically throws the following error: 
`ERROR:tmux:WeatherSegment:Exception while computing state for None: u'condition'`

The relevant segment definition: 
` .config/powerline/themes/tmux/custom.json`:
...
```JSON
      {
        "function": "powerline.segments.common.wthr.weather",
        "args": {
        "interval": 1000
        }
      }
```
...
Then after some time it updates the segment just fine. 

Also, on a tangent, how about a adding [forecast.io](https://developer.forecast.io/docs/v2) as a weather provider?

PR #1681, opened 2016-11-13
retry 5 times before throwing an exception

rchicoli:
after upgrading my system to latest I had some issue with the weather segment.  I was getting sporadic errors:
```ERROR:tmux:WeatherSegment:Exception while computing state for None: 'NoneType' object has no attribute ' getitem '```
this happens because some times `response['query']['results'] is None`  
so I added at least a retry up to 5 times.  
problem solved for me: this is related to: #1119

### DIFF
--- a/powerline/segments/common/wthr.py
+++ b/powerline/segments/common/wthr.py
@@ -136,13 +136,20 @@ class WeatherSegment(KwThreadedSegment):
 			return url
 
 	def compute_state(self, location_query):
-		url = self.get_request_url(location_query)
-		raw_response = urllib_read(url)
-		if not raw_response:
-			self.error('Failed to get response')
-			return None
+		for i in range(5):
+			try:
+				url = self.get_request_url(location_query)
+				raw_response = urllib_read(url)
+				if not raw_response:
+					self.error('Failed to get response')
+					return None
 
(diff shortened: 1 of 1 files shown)


rchicoli:
this is not failing because of my changes. the master and the develop branch is broken as you can see it here https://travis-ci.org/rchicoli/powerline/builds/175796522

rchicoli:
I was checking this today again. It seems that this problem only happened that day. So I guess this issue happened because the endpoint `query.yahooapis.com` was returning some wrong payload time to time.
Even though I kinda like to have a retry option, it doesn't hurt and helped for that time.
You can choose either to accept this change or feel free to close it, because the endpoint is healthy again (returning good date).

rchicoli:
I will create another MR for fixing the checks with travis, because they are not working. So we have different issues, not related to each other.

rchicoli:
exactly what I though, after solving this issue https://github.com/powerline/powerline/issues/1119, now everything works fine.
I still like the idea of the retry though, but this would be necessary only if you request the endpoint every second, because some requests timeout.
So I will close this issue for now.

rchicoli:
feel free to delete the branch if you don't want to merge the changes

PR #1694, opened 2016-12-08
Add file name to dict in mpd.

### LINKED ISSUE #1693: Mpd player segment Show filename instead of album/artist
My music collection is mostly mp3's ripped lazily. I have them organized in directories as "artist"/"song title.mp3".  Very few of them have the album info correctly and few of them are outright wrong. 

Mpc can return the file name, starting from the root of the default music directory using '%file%' format tag. This is exactly what I need for displaying the song title and artist. Usually it's "None - None (5:13)" for me.

BlitzKraft:
Fixes #1693 

Now the player mpd segment can also display filename.

### DIFF
--- a/powerline/segments/common/players.py
+++ b/powerline/segments/common/players.py
@@ -101,6 +101,10 @@ Highlight groups used: ``player_fallback`` or ``player``, ``player_play`` or ``p
 	+------------+-------------------------------------------------------------+
 	|total       |Composition length in format M:SS.                           |
 	+------------+-------------------------------------------------------------+
+	|file        |Mpd only option. Show file name of the song currently being  |
+	|            |played with path. Path shown is relative to the default      |
+	|            |mpd music directory.                                         |
+	+------------+-------------------------------------------------------------+
 :param dict state_symbols:
 	Symbols used for displaying state. Must contain all of the following keys:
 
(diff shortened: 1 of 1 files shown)


ZyX-I:
Missing documentation update.

BlitzKraft:
@ZyX-I 

I don't see where I can update the docs. The docs folder is where I can update the documentation, right?

ZyX-I:
No, documentation for segments is in source code. It is common for all player segments.

And you also missed an update to PlayerSegment class.

ZyX-I:
(Documentation is below PlayerSegment class. Missed default value for the new key and documentation for it.)

BlitzKraft:
Made all the changes. Added in documentation and added default value in the player segment class.

powerline/segments/common/players.py
```diff
@@ -101,6 +102,10 @@ def omitted_args(self, name, method):
 	+------------+-------------------------------------------------------------+
 	|total       |Composition length in format M:SS.                           |
 	+------------+-------------------------------------------------------------+
+	|file        |Mpd only option. Show file name of the song currently being  |
```
ZyX-I: I do not see why this must be MPD only, and this is not an option. Should be something like “(!No ‘show’, this is a parameter which exists, it cannot do any actions) File name of the song currently being played.” Plus some note in mpd documentation that “file” parameter is set to the file name with path relative to the default MPD music directory.

PR #1704, opened 2017-01-13
Installation for Windows support

vincpa:
Added documentation for Windows support with the only Python 2.7 as the test version.

### DIFF
--- a/docs/source/installation/windows.rst
+++ b/docs/source/installation/windows.rst
@@ -1,0 +1,43 @@
+********************
+Installation on Windows
+********************
+
+Python package
+==============
+
+1. `Install Python <https://www.python.org/downloads>`_ version 2.7.12, 2.8 or 3.4 upwards - Python 2.7.12 onwards, contains pip in the installation package.
+
+2. Install Powerline using one of the following commands:
+
+   .. code-block:: ps1
+
+       pip install --user powerline-status
+
+   will get current release version.
+
+   .. note::
+      Powerline developers should be aware that ``pip install --editable`` does 
+      not currently fully work. Installation performed this way are missing 
+      ``powerline`` executable that needs to be symlinked. It will be located in 
+      ``scripts/powerline``.
+
+Vim installation
+================
+
+Before you begin the installation, ensure that your Vim and Python supported CPU architecture are either both 32bit or both 64bit. Python support in Vim will not work if they are different.
+
+You'll also need to ensure you have python27.dll in either your vim74 or vim80 installation directory (that is if you plan on using Python 2.7)
+
+Clone the powerline git repository. The only file you'll need from this repository is
+
+   .. code-block:: sh
+
+       bindings\vim\plugin\powerline.vim
+
+Depending on your version of Vim, copy that file to either vim74\plugin\powerline.vim or vim80\plugin\powerline.vim
+
+
+Fonts installation
+==================
+
+To install the font, copy the font PowerlineSymbols.otf from the powerline git repository to vimfiles\fonts, which can be found in your vim installation directory.


docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
+********************
+Installation on Windows
+********************
```
ZyX-I: Amount of asterisks must be greater or equal to the title.

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
...
+   .. note::
+      Powerline developers should be aware that ``pip install --editable`` does 
+      not currently fully work. Installation performed this way are missing 
+      ``powerline`` executable that needs to be symlinked. It will be located in 
+      ``scripts/powerline``.
+
+Vim installation
+================
+
+Before you begin the installation, ensure that your Vim and Python supported CPU architecture are either both 32bit or both 64bit. Python support in Vim will not work if they are different.
```
ZyX-I: Lines are too long. Documentation must wrap at 80’th column unless it is a code block.

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
...
+
+You'll also need to ensure you have python27.dll in either your vim74 or vim80 installation directory (that is if you plan on using Python 2.7)
+
+Clone the powerline git repository. The only file you'll need from this repository is
+
+   .. code-block:: sh
+
+       bindings\vim\plugin\powerline.vim
```
ZyX-I: There is no reason to put this file in a code block, inline it.

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Depending on your version of Vim, copy that file to either vim74\plugin\powerline.vim or vim80\plugin\powerline.vim
```
ZyX-I: File names are normally wrapped with <code>:file:\`…\`</code> in documentation (also applies to the file name from the code block because it should be moved out of it).

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
...
+
+       pip install --user powerline-status
+
+   will get current release version.
+
+   .. note::
+      Powerline developers should be aware that ``pip install --editable`` does 
+      not currently fully work. Installation performed this way are missing 
+      ``powerline`` executable that needs to be symlinked. It will be located in 
+      ``scripts/powerline``.
```
ZyX-I: Remove note, `powerline` executable is not supposed to work on Windows (none of the variants), so it does not matter whether or not pip will install it.

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
+********************
+Installation on Windows
+********************
+
+Python package
+==============
+
+1. `Install Python <https://www.python.org/downloads>`_ version 2.7.12, 2.8 or 3.4 upwards - Python 2.7.12 onwards, contains pip in the installation package.
+
+2. Install Powerline using one of the following commands:
```
ZyX-I: There is only one following command, not many. “One of the following” is not applicable.

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
...
+      not currently fully work. Installation performed this way are missing 
+      ``powerline`` executable that needs to be symlinked. It will be located in 
+      ``scripts/powerline``.
+
+Vim installation
+================
+
+Before you begin the installation, ensure that your Vim and Python supported CPU architecture are either both 32bit or both 64bit. Python support in Vim will not work if they are different.
+
+You'll also need to ensure you have python27.dll in either your vim74 or vim80 installation directory (that is if you plan on using Python 2.7)
```
ZyX-I: You need to be sure that dll is somewhere in $PATH\*, or in the same directory as the Vim executable AFAIK. Vim executable currently happens to be in `…/vim{version}`, but this directory is not special otherwise.

\*  May be edited in registry in `HKLM\System\CurrentControlSet\Control\Session Manager\Environment`: at least this is how I edit `$PATH` in wine.

docs/source/installation/windows.rst
ZyX-I: And I will not accept any installation advices which suggest to make any changes to Vim installation directory. This includes suggestion in this line and suggestion regarding Python dll. Properly done installation is immune to Vim reinstalls, even if Vim installer happens to clear installation directory before proceeding (I do not really know whether or not it does, but even if it does not version switches (7.4 to 8.0) have similar effect because installation directory changes).

docs/source/installation/windows.rst
```diff
@@ -0,0 +1,43 @@
+********************
+Installation on Windows
+********************
+
+Python package
+==============
+
+1. `Install Python <https://www.python.org/downloads>`_ version 2.7.12, 2.8 or 3.4 upwards - Python 2.7.12 onwards, contains pip in the installation package.
```
ZyX-I: Where did you see version 2.8? Also do not hardcode patch version number, it may change.

ZyX-I: Also where is the reference that Python version must be the same as the Python version used to compile Vim (at least, major+minor, but better if major+minor+patch)?

docs/source/installation/windows.rst
ZyX-I: *Especially* advices to ***copy*** anything anywhere like here: proper installation should use files installed by pip/Python installer so that copying procedure does not need to be redone when updating powerline/Python.

docs/source/installation/windows.rst
vincpa: When Python is unavailable and Vim starts up, powerline presents a popup, it specifically states that those versions are required for powerline.

docs/source/installation/windows.rst
vincpa: Because there is no "official" support for Windows, I have to resort to jumping through hoops to figure this stuff out. Usually plugins install nicely using pathogen, but powerline doesn't on Windows.

docs/source/installation/windows.rst
ZyX-I: @vincpa Screenshot? There are exactly no mentions of `2.8` anywhere in the repository, except for seven floating-point constants in tests/test_segments.py (like `62.857142857142854`).

docs/source/installation/windows.rst
ZyX-I: Instructions for enabling powerline on Windows are generally the same as for other systems, except that getting Vim with Python support is more tricky then just “make package manager install Vim with Python support for you, it will pull Python and pip as dependencies”: this could be the most useful part of this file.

Documentation for Vim suggests that you either add `set runtimepath+=…` to your vimrc or do not use `powerline.vim` at all, running a few `:py` commands instead. Both variants do not require copying `powerline.vim` file anywhere and first variant should make Vim source powerline without copying powerline file.

Issue #1688, opened 2016-11-23
Not visible in Vim unless multiple windows are open


qwertystop:
Title says it all. Powerline is normally non-present (The status line with -- INSERT -- or the last typed command is still there), but when I have multiple windows open (whether vertically or horizontally split, and whether it's multiples of the same buffer or multiple buffers), it shows up under all of them as it should.

I can provide configuration information as necessary but don't want to just dump a wall of everything on you. This is a terminal Vim, not running in a GUI; version is 7.4, compiled with Python 3 support. OS is Mint. I vaguely recall an issue like this a while ago when I was setting up on a previous system, but can't seem to find anything on it now, so that's just a vague assurance that it's probably solvable?

ZyX-I:
This is not a bug, `:h 'laststatus'`.

bennypowers:
This should be added to the docs.

qwertystop:
Agreed there. Default behavior, not obviously intended functionality, not obviously configurable, and you only need to do it once per Vim install so it's easy to forget about by the next time you need it.

PR #1706, opened 2017-01-21
Added note about laststatus

bennypowers:
Users may be confused when installing powerline as it does not appear by default in single-window mode. This change alerts users that they should not expect to see the status line and points them to the help documents for laststatus so that they can activate it in all cases if they want.

See #1688

### DIFF
--- a/docs/source/usage/other.rst
+++ b/docs/source/usage/other.rst
@@ -17,6 +17,9 @@ If installed using pip just add
 
 (replace ``python`` with ``python3`` if appropriate) to the :file:`vimrc`.
 
+.. note::
+    Status line will not appear by default when there is only a single window displayed. `:h laststatus` for more information.
+
 If the repository was just cloned the following line needs to be added to the 
 :file:`vimrc`:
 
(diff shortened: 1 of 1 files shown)


docs/source/usage/other.rst
```diff
@@ -28,8 +31,7 @@ where ``{repository_root}`` is the absolute path to the Powerline installation
 directory (see :ref:`repository root <repository-root>`).
 
 If pathogen is used and Powerline functionality is not needed outside of Vim 
-then it is possible to simply add Powerline as a bundle and point the path above 
-to the Powerline bundle directory, e.g. 
+then it is possible to simply add Powerline as a bundle and point the path above to the Powerline bundle directory, e.g.
```
ZyX-I: Unrelated, too long line. Must not be here.

docs/source/usage/other.rst
```diff
@@ -17,6 +17,9 @@ If installed using pip just add
 
 (replace ``python`` with ``python3`` if appropriate) to the :file:`vimrc`.
 
+.. note::
+    Status line will not appear by default when there is only a single window displayed. `:h laststatus` for more information.
```
ZyX-I: Tag is named `'laststatus'` (with single quotes), there is no `laststatus` tag: it works only because Vim also does substring matching. Code should be enclosed in *two* backticks, not one. Line too long, wrap at 80’th column.

Also many other notes are aligned with `n` in `note`, though not all AFAIK, can’t say what is more common. I align with `n` when I remember to align with `n`.

// And information is already in troubleshooting, third Vim issue.

docs/source/usage/other.rst
```diff
@@ -17,6 +17,10 @@ If installed using pip just add
 
 (replace ``python`` with ``python3`` if appropriate) to the :file:`vimrc`.
 
+.. note::
+   Status line will not appear by default when there is only a single window
+   displayed. `:h 'laststatus'` for more information.
```
ZyX-I: Double backticks. Missing verb in the second sentence.

PR #1717, opened 2017-02-06
Fish bindings: replace eval with source

teozkr:
For some reason, fish 2.5.0 freaks out under certain circumstances if
your `config.fish` calls `eval` at any point. For equally mysterious
reasons, replacing `eval` with `source` makes things work again.

See fish-shell/fish-shell#3809 for more details.


teozkr:
Sorry, that should never have worked in the first place, because Fish doesn't support heredocs. I have no idea why it did work for me until I rebooted.

Anyway, replaced it with echo | source now, which should work and doesn't rely on features that Fish doesn't have.

PR #1720, opened 2017-02-13
Make length of `output_padding` in renderer controllable

welwood08:
Defaults to previous behaviour (length 1).

I'm not sure if anyone else would be interested in this change, but the compulsory padding was bugging me so here it is.

### DIFF
--- a/powerline/renderer.py
+++ b/powerline/renderer.py
@@ -458,7 +458,7 @@ class Renderer(object):
 					segment is first_segment
 					if side == 'left' else
 					segment is last_segment
-				))
+				)) * theme.outer_padding
 
 				draw_divider = segment['draw_' + divider_type + '_divider']
 				segment_len += outer_padding
@@ -519,7 +519,7 @@ class Renderer(object):
 					segment is first_segment
 					if side == 'left' else
 					segment is last_segment
-				)) * ' '
+				)) * theme.outer_padding * ' '
 				divider_type = 'soft' if compare_segment['highlight']['bg'] == segment['highlight']['bg'] else 'hard'
 
 				divider_highlighted = ''
--- a/docs/source/configuration/reference.rst
+++ b/docs/source/configuration/reference.rst
@@ -361,6 +361,10 @@ ascii                       Theme without any unicode characters at all
        is set in the local themes it will be ignored. This option may also be 
        ignored in some bindings.
 
+``outer_padding``
+    Defines number of spaces at the end of output (on the right side) or at 
+    the start of output (on the left side). Defaults to ``1``.
+
 
 ``dividers``
     Defines the dividers used in all Powerline extensions.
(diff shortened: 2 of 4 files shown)


powerline/renderer.py
```diff
@@ -458,7 +458,7 @@ def _render_length(self, theme, segments, divider_widths):
 					segment is first_segment
 					if side == 'left' else
 					segment is last_segment
-				))
+				)) * int(self.theme_config.get('outer_padding', 1))
```
ZyX-I: This is not the right place for defaults (especially given that you have to repeat default value twice). Mostly defaults are placed in `powerline/__init__.vim`, though it looks like theme_config currently does not allow omitting most things and rare defaults are not there. Also note that `theme_config` attribute is not used anywhere, but for defining `Theme` instances: you should take necessary value from theme, first saving it there with something like

```patch
diff --git a/powerline/theme.py b/powerline/theme.py
index 26b67a3..91ea3a2 100644
--- a/powerline/theme.py
+++ b/powerline/theme.py
@@ -69,6 +69,7 @@ class Theme(object):
 			self.cursor_space_multiplier = None
 		self.cursor_columns = theme_config.get('cursor_columns')
 		self.spaces = theme_config['spaces']
+		self.outer_padding = theme_config.get('outer_padding', 1)
 		self.segments = []
 		self.EMPTY_SEGMENT = {
 			'contents': None,
```

welwood08:
Thanks for the pointer, commit amended and force-pushed to branch.

I don't often write python code, so let me know if there should be a getter in the `Theme` for the new property or if a test is needed.

ZyX-I:
@welwood08 Yes, test would be good. (Current suite fails, but there is a nice summary below which tells which parts of the suite fail: normally it should be shell and tmux.) Something like this:

```Patch
diff --git a/tests/test_configuration.py b/tests/test_configuration.py
index 56ca0c2..7bec000 100644
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -550,6 +550,23 @@ class TestDisplayCondition(TestRender):
 			self.assertRenderEqual(p, '{56} s1{6-}>>{--}', mode='m1')
 
 
+class TestOuterPadding(TestRender):
+	@add_args
+	def test_outer_padding_zero(self, p, config):
+		config['themes/' + UT]['outer_padding'] = 0
+		self.assertRenderEqual(p, '<RESULT HERE>', width=50)
+
+	@add_args
+	def test_outer_padding_ten(self, p, config):
+		config['themes/' + UT]['outer_padding'] = 10
+		self.assertRenderEqual(p, '<RESULT HERE>', width=50)
+
+	@add_args
+	def test_outer_padding_left(self, p, config):
+		config['themes/' + UT]['outer_padding'] = 5
+		self.assertRenderEqual(p, '<RESULT HERE>', width=50, side='left')
+
+
 class TestSegmentAttributes(TestRender):
 	@add_args
 	def test_no_attributes(self, p, config):
```

Can be checked locally with `python tests/test_configuration.py -v`.

ZyX-I:
Though you may also need to edit (like editing `'themes/' + UT`, not `config` global variable itself) `config['themes/test/default']`.

welwood08:
Apologies for not getting back to this sooner, I've now added tests based on your example and they pass when tested locally.

PR #1729, opened 2017-02-25
Prevent divide by zero battery capacity

thomwiggers:
If your battery is broken this segment crashes with a ZeroDivisionError


ZyX-I:
Missing a test for this.

ZyX-I:
Though no, this part is not tested.

PR #1732, opened 2017-03-03
Itunes players/ music player change?

chester755:
From the old version of the powerline, I have the ITunes player segment fixed. I have also change the way how the player segment looks, now you can see how far down the music has been played.


powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
 
 {0}
 ''').format(_common_args.format('rdio')))
+
+
+
```
ZyX-I: Too much empty lines.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
+class iTunesPlayerSegment(PlayerSegment):
```
ZyX-I: Class names must start with a capital letter.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
+        def get_player_status(self, pl):
+                status_delimiter = '-~`/='
+                ascript = '''
+                tell application "System Events"
+                set process_list to (name of every process)
```
ZyX-I: Missing indentation. Script must be readable.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
 
 {0}
 ''').format(_common_args.format('rdio')))
+
+
+
+class iTunesPlayerSegment(PlayerSegment):
+        def get_player_status(self, pl):
```
ZyX-I: Do not use spaces for indentation, powerline uses tabs.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
+                else:
+                        title
+                if len(artist) > 10:
+                        artist = artist[0:10] + "..."
+                else:
+                        artist
+                if len(album) > 10:
+                        album = album[0:10] + "..."
+                else:
+                        album
```
ZyX-I: First: all `else` clauses are completely unneeded, you do not do *anything* at all here. Second: if you need truncation, truncate in format string. Truncations from here must be removed completely, also with preceding comment.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
+                else:
+                        artist
+                if len(album) > 10:
+                        album = album[0:10] + "..."
+                else:
+                        album                        
+                state = _convert_state(now_playing[5])
+                total = _convert_seconds(now_playing[4])
+                elapsed = _convert_seconds(float(now_playing[3]))
+                #elapsed = _convert_seconds(float(now_playing[4]) - float(now_playing[4]))
```
ZyX-I: This comment also should be removed, code in it does not work.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
+                        'title': title,
+                        'artist': artist,
+                        'album': album,
+                        'total': total,
+                        'elapsed': elapsed,
+                        'state': state
+                }
+        
+itunes = with_docstring(iTunesPlayerSegment(),
+('''Return iTunes now playing information.
```
ZyX-I: Missing newline after summary. Also I normally do *not* end summaries with dot.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,70 @@ def get_player_status(self, pl):
...
 {0}
 ''').format(_common_args.format('rdio')))
+
+
+
+class iTunesPlayerSegment(PlayerSegment):
+        def get_player_status(self, pl):
+                status_delimiter = '-~`/='
+                ascript = '''
+                tell application "System Events"
```
ZyX-I: And additionally inner data should be more indented.

powerline/segments/common/players.py
```diff
@@ -34,7 +34,7 @@ def _convert_seconds(seconds):
 
 
 class PlayerSegment(Segment):
-	def __call__(self, format='{state_symbol} {artist} - {title} ({total})', state_symbols=STATE_SYMBOLS, **kwargs):
+	def __call__(self, format='{state_symbol}|{artist} - {title}|({elapsed}/{total})', state_symbols=STATE_SYMBOLS, **kwargs):
```
ZyX-I: Pipes should be spaces back, this is more in line with powerline style (though having state/artist/title/(elapsed/total) in separate segments would be even more in line, but doing this properly would be far harder).

chester755:
Thanks for all the comments, I think I have changed everything and recovered the origin of this change. I do agree that adding this elapsed will get more in line. But this is how this looks on my tmux.
screenshot 2017-03-06 10 48 51

powerline/segments/common/players.py
```diff
@@ -477,3 +477,57 @@ def get_player_status(self, pl):
 
 {0}
 ''').format(_common_args.format('rdio')))
+
+
+class ITunesPlayerSegment(PlayerSegment):
+	def get_player_status(self, pl):
+		status_delimiter = '-~`/='
+		ascript = '''
+		tell application "System Events"
```
ZyX-I: Internal script is still not indented.

ZyX-I:
@chester755 You can add truncation to the format string, Python format allows this. Also I am not against elapsed change there, just don’t use pipes. And do not forget that this argument may be overridden by configuration (including adding truncation there, though I do not know a way to make Python put `…` when it occurs).

chester755:
Again, thanks for all the suggestions, after checking all the other music players, I decide to remove formatting string part. I think it will be better for users to change that in config file. Emacs is a pain sometimes......

ZyX-I:
.DS_store file should not be there.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,54 @@ def get_player_status(self, pl):
...
+		set t_elapsed to player position
+		set t_state to player state
+		return t_title & "{0}" & t_artist & "{0}" & t_album & "{0}" & t_elapsed & "{0}" & t_duration & "{0}" & t_state
+		end if
+		end tell
+		end if
+		'''.format(status_delimiter)
+		now_playing = asrun(pl, ascript)
+		if not now_playing:
+		return
```
ZyX-I: Looks like SyntaxError: missing indent.

chester755:
mac is annoying.....

chester755:
I think it should be good now. Again thanks for all the checks and suggestions.

ZyX-I:
Missing indentation of the ascript: check examples above.

chester755:
I think it should be fixed now.

powerline/segments/common/players.py
```diff
@@ -477,3 +477,54 @@ def get_player_status(self, pl):
...
+		status_delimiter = '-~`/='
+		ascript = '''
+			tell application "System Events"
+				set process_list to (name of every process)
+			end tell
+		
+			if process_list contains "iTunes" then
+				tell application "iTunes"
+					if player state is playing then
+						set t_title to name of current track
```
ZyX-I: By the way, one of the above scripts uses `set t_title to the name of the current track`: two additional `the`s. Does this have special meaning?

chester755: I am not sure.... Again I believe this exist in the previous Powerline segments or somewhere online I can't recall. I took most part of it and modified some parts to make it work.

Issue #1727, opened 2017-02-22
tmux master branch version issue


Chris-Bee:
Hello,
when compiling the tmux from source (master branch). tmux -V shows "master" as the version number. This causes "source .powerline/powerline/bindings/tmux/powerline.conf" to fail. If I checkout the latest tag of tmux so that tmux -V shows a version number e.g. 2.3 - "source .powerline/powerline/bindings/tmux/powerline.conf" works fine.

BlitzKraft:
I had the same issue. when I checked out master. I changed the file in `bindings/tmux/__init__.py` to hard code the version info for now. 

```
Line 76: version_string = "tmux 2.3"
```

This fixed it. Excepting that, I don't think there is much that can be done from powerline.

s-ol:
@BlitzKraft: well, sane behaviour would be to expect tmux to be of the latest supported version if it reports `master`.

Chris-Bee:
@s-ol That makes the most sense. Can anybody do it and just do a pull request?

PR #1735, opened 2017-03-07
Assume latest tmux version if -V returns 'master'

s-ol:
See #1727.

It seems that [tmux/tmux](https://github.com/tmux/tmux/branches) is only developed on a single `master` branch, so having that comparison hardcoded is probably appropriate, but hardcoding the latest tmux version to be bumped every release seems less-than-ideal to me, open to suggestions on that.

Since tmux releases may prompt changes to the tmux bindings anyhow though, it  may be 'good enough' but I suspect something better should be done in `powerline/bindings/config.py` to snap to the latest version based on the available tmux.conf's in the binding directory.

### DIFF
--- a/powerline/bindings/tmux/__init__.py
+++ b/powerline/bindings/tmux/__init__.py
@@ -76,6 +76,8 @@ def get_tmux_version(pl):
 	version_string = get_tmux_output(pl, '-V')
 	_, version_string = version_string.split(' ')
 	version_string = version_string.strip()
+	if version_string == 'master':
+		return TmuxVersionInfo(float('inf'), 0, version_string)
 	major, minor = version_string.split('.')
 	suffix = DIGITS.subn('', minor)[0] or None
 	minor = NON_DIGITS.subn('', minor)[0]


ZyX-I:
I would say it is logical to supply infinity there. Why do you need exact version? If it was C I would suggest `{TYPE}_MAX`, but there is no requirement to have integers.

s-ol:
ah, I didnt really think of that.

powerline/bindings/tmux/__init__.py
```diff
@@ -76,6 +76,8 @@ def get_tmux_version(pl):
 	version_string = get_tmux_output(pl, '-V')
 	_, version_string = version_string.split(' ')
 	version_string = version_string.strip()
+	if version_string == 'master':
+		return TmuxVersionInfo(math.inf, 0, version_string)
```
ZyX-I: And where is the `math` import? Also `math` module does not have `inf` attribute. Just use `float('inf')`, your variant is completely broken.

s-ol: Woah, sorry for that. Apparently I accidentally tested this with my 2.3-tagged version twice instead of using the `master` build.

`math.inf` definetely exists in my python 3.6 interpreter, but i changed it to `float('inf')` anyway.

ZyX-I: I use 3.4 by default. And powerline supports 2.6+ and 3.2+, so there are lots of supported versions which dou not have `math.inf`.

PR #1740, opened 2017-03-14
right prompts for bash

mjg:
Provide a right prompt for bash so that with the default config, so
that those right segments don't fall into oblivion.

Given the way they are output, there is no adjustment of the current
line for window width cahnhes, of course, but all subsequent lines will
be adjusted to the new prompt.

### DIFF
--- a/powerline/bindings/bash/powerline.sh
+++ b/powerline/bindings/bash/powerline.sh
@@ -67,6 +67,12 @@ _powerline_prompt() {
 _powerline_set_prompt() {
 	local last_exit_code=$?
 	local jobnum="$(jobs -p|wc -l)"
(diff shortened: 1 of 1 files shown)


ZyX-I:
Not merging this:

1. Does not work with `above` lines: printing happens before prompt is shown, so it will be placed on the uppermost line.
2. Is not redrawn: type `<C-l>` and this “right” prompt is gone.
3. Prints to stdout, other prompts bash prints to stderr.
4. Right prompt should disappear if user types long enough command line, not look like part of his input.
5. Right prompt must not disappear if user happens to delete something.
6. Right prompt must reappear if user input is no longer lengthy.

It is better to not have any right prompt at all then to have such a bad right prompt. Bash does not have enough features to emulate right prompt on top of anything it has.

PR #1760, opened 2017-04-29
Fix tmux vterm tests

ZyX-I:
First commits originate from #1507.


ZyX-I:
This reenables part of the tests disabled by #1723. Other part will unlikely be reenabled, rather then moved to be vterm-based.

PR #1765, opened 2017-05-06
add Terminus alias

evgeni:
this allows using Terminus with PowerlineSymbols in xfce4-terminal and similar terminals


ZyX-I:
Terminus now has powerline symbols merged in: see [4.39 changelog](http://terminus-font.sourceforge.net/). You need to update your font.

evgeni:
Thanks. It does indeed work on my Fedora box, but not on my Debian box. Despite both having Terminus 4.40. Will investigate further, but that's obviously not powerline related anymore :)

PR #1771, opened 2017-05-13
Replace awesome-client with a Python-native dbus interface.

brycied00d:
Instead of spawning awesome-client every couple of seconds to update the
widget, use Python's own native dbus interface. Fewer system resources used by
avoiding the spawn as well as removing noise on the dbus caused by open+close.

I haven't updated the tests for this as I'm unsure just how to test. The current tests
replace `awesome-client` with its own version that logs the request but doesn't
connect to awesome. Problem is that by removing that layer(s), I haven't a clue
how to "intercept" or shim the dbus interface. I'm happy to take suggestions.

### DIFF
--- a/powerline/bindings/wm/awesome.py
+++ b/powerline/bindings/wm/awesome.py
@@ -1,25 +1,12 @@
 # vim:fileencoding=utf-8:noet
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
-import sys
-
 from threading import Thread, Event
 from time import sleep
-from subprocess import Popen, PIPE
+import dbus
 
 from powerline import Powerline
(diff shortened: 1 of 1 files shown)


ZyX-I:
I guess you will have to remove `tests/test_awesome` and use `tests/test_python` with dbus module mocks (search for `replace_module_module` in `tests/test_python/test_….py` files). Though if you have some other idea which includes running tests with awesome and dependencies being installed, but not running (so far there are no such tests for WM bindings because I had no such ideas) then you are also welcome.

brycied00d:
Thanks for the tip, @ZyX-I I'll get to work on implementing that, fixing the CI failure and re-adding `awesome-client` as a fallback when `python-dbus` isn't installed.

PR #1776, opened 2017-05-17
✏️ Add fish to list of supported apps on readme

phatblat:


s-ol:
shouldn't there have been a space at the end of that line?

PR #1782, opened 2017-06-02
Fix for "last_pipe_status: bad array subscript"

lstrojny:
Recent versions of powerline lead to this error in bash: 

```
-bash: last_pipe_status: bad array subscript
```

### DIFF
--- a/powerline/bindings/bash/powerline.sh
+++ b/powerline/bindings/bash/powerline.sh
@@ -34,7 +34,7 @@ _powerline_return() {
 _powerline_status_wrapper() {
 	local last_exit_code=$? last_pipe_status=( "${PIPESTATUS[@]}" )
 
-	if test "$last_exit_code" != "${last_pipe_status[-1]}" ; then
+	if test "$last_exit_code" != "${last_pipe_status[0]}" ; then
 		last_pipe_status=()
 	fi
 	"$@" $last_exit_code "${last_pipe_status[*]}"


ZyX-I:
This “fix” is entirely incorrect. `$?` should be equal to the status of the last program in pipe, not to the first.

BTW, which versions? I have 4.3_p48-r1 and it is working fine here: `-1` subscript means last array element.

lstrojny:
I have 3.2.57

lstrojny:
Upgraded to 4.4.12, didn't help

lstrojny:
Btw, interestingly enough, the change works just fine 
1__bash

ZyX-I:
@lstrojny Where do you see a pipe in the example? Of course if `$PIPESTATUS` has only *one* element then “last” element is the same thing as “first”.

ZyX-I:
And see the tests, you have failed all of them due to this exact problem: missing `42` in the diff is because `$?` is equal to `43`, but first element of a pipe is `42`:

```
-  BRANCH  …  shell  3rd  «Unicode!»  42  43  set_theme_option default_leftonly.segments.above "$ABOVE_LEFT"
+  BRANCH  …  shell  3rd  «Unicode!»  43  set_theme_option default_leftonly.segments.above "$ABOVE_LEFT"
```

ZyX-I:
Bash version on travis is 4.2-2ubuntu2.6.

ZyX-I:
If you see the same behaviour after upgrading maybe then either you have not restarted your shell after upgrade or it is controlled by some option. Though it is zsh which is mainly famous for creating user options that break scripts (up to an option which controls whether array subscripting is going to start from one or zero which obviously makes array subscripts impossible to use without starting your function with `emulate -L zsh`), have not heard of bash ones (`set -e`/`set -u`/etc are hardly *user* options, they are for scripts unless user is seeking for troubles).

ZyX-I:
Also note that even if you remove `if` and `fi` lines entirely powerline is not going to fail drastically: just when you used a pipe you will no longer see more then one status. So do not perform any tests which are not showing more then one status yourself, better check `true | false` which should show you something like



(with arrows in different direction, of course). If you get only `1` out of `true | false` then you have broken the code.

lstrojny:
Thanks @ZyX-I that makes all a lot of sense. Let me see if I can find anything and sure, the patch doesn’t make any sense as-is.

ZyX-I:
You may need to use array length (`${#arrname[@]}`) minus one as a replacement, but since I don’t know why current code fails I could not suggest a real fix.

lstrojny:
The last_pipe_status array is empty, so I am guessing the root cause is happening somewhere else

ZyX-I:
`$PIPESTATUS` variable got disabled somehow? Zsh is also known to provide a script author lots of ways to mess with its builtins (disable via some options, make normally special variables locally not special, etc). (I have a better knowledge of how to mess things up when using zsh then how to mess things up when using bash.)

lstrojny:
OK, for further reference, it was the old bash version. Somehow iTerm didn’t pick up the new version in /usr/local/bin. Properly upgrading to 4.4.12 helped.

lstrojny:
Thanks @ZyX-I !

ZyX-I:
@lstrojny If your bash version is not uncommon I would not drop supporting it. Just you need a different fix: add a condition which checks whether array in question is not empty. (Better actually check whether `$PIPESTATUS` is special though in case somebody defined `$PIPESTATUS` for some reason: e.g. like writing `local PIPESTATUS=foo; if test "$PIPESTATUS" != "foo"`, except that I remember seeing something like “fixed a crash when assigning the pipe status” in a changelog so this is not an option.)

pEdler:
When you say "properly upgrading to 4.x.x" what do you mean? I'm having the same issues on OSX with gnu bash 4.4

lstrojny:
@pEdler I am using iTerm2 on OS X and homebrew to install a newer bash version. The newer bash version installed by homebrew is installed in `/usr/local/bin`, while OS X stock version is installed at `/bin`. Even though the precedence of the `PATH` variable puts `/usr/local/bin`before `/bin`, iTerm 2 didn’t respect that path variable and always used the old bash version. I've changed the profile settings in iTerm 2 under `Profiles -> <profile name> -> General -> Command` to use `/usr/local/bin/bash` as a Command instead of the default which is "Login shell"

taitung:
faced the same err of "bad array subscript" and checked https://github.com/powerline/powerline/blob/develop/powerline/bindings/bash/powerline.sh at line 48 and found a hashtag before the array, which is probably the root cause. I removed it and it works now. 
`| test "${**#**last_pipe_status[@]}" -eq "0" \`

RamaDL:
Thanks! I solved the problem upgrading BASH with MacPorts:
`sudo port install bash`
And changing the absolute path where terminal.app bring the bash executable, I did this enabling Command(full path) on menu Terminal->Preferences->General->Command(full path) and filling de field with: 
`/opt/local/bin/bash`.

PR #1803, opened 2017-07-13
Fix powerline with gcc7

cryptomilk:
Hi,

the first two patches fix the compilation of powerline with GCC 7.x. The last one is a small optimization.

### DIFF
--- a/client/powerline.c
+++ b/client/powerline.c
@@ -71,7 +71,7 @@ inline size_t true_sun_len(const struct sockaddr_un *ptr) {
 #endif
 
 #define ADDRESS_SIZE sizeof(ADDRESS_TEMPLATE) + (sizeof(uid_t) * 4)
-#define NUM_ARGS_SIZE (sizeof(int) * 2)
+#define NUM_ARGS_SIZE (sizeof(int) * 2 + 1)
 #define BUF_SIZE 4096
 #define NEW_ARGV_SIZE 200
 
--- a/client/powerline.c
+++ b/client/powerline.c
@@ -42,7 +42,7 @@ void do_write(int sd, const char *raw, size_t len) {
 	}
 }
 
-inline size_t true_sun_len(const struct sockaddr_un *ptr) {
+static inline size_t true_sun_len(const struct sockaddr_un *ptr) {
 #ifdef __linux__
 	/* Because SUN_LEN uses strlen and abstract namespace paths begin
 	 * with a null byte, SUN_LEN is broken for these. Passing the full
--- a/client/powerline.c
+++ b/client/powerline.c
@@ -88,6 +88,7 @@ int main(int argc, char *argv[]) {
 	char *wd = NULL;
 	char **envp;
 	const char *address;
+	int len;
 
 	if (argc < 2) {
 		printf("Must provide at least one argument.\n");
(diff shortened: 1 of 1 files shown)


client/powerline.c
```diff
@@ -42,7 +42,7 @@ void do_write(int sd, const char *raw, size_t len) {
 	}
 }
 
-inline size_t true_sun_len(const struct sockaddr_un *ptr) {
```
ZyX-I: What’s the problem with `inline`? And does not this work with `static inline`? This looks like a gcc bug so far.

client/powerline.c
```diff
@@ -88,6 +88,7 @@ int main(int argc, char *argv[]) {
 	char *wd = NULL;
 	char **envp;
 	const char *address;
+	size_t len;
```
ZyX-I: `snprintf` officially returns `int`.

cryptomilk:
Yes, it should be static inline.

ZyX-I:
Some notes:

1. `s[n]printf` will *always* terminate the string. What warning was complaining about is not absence of string termination, but that should the number have the INT_MAX order of magnitude (actually with `%x` UINT_MAX, but we have an int) it will be *truncated*. Given that system limits number of arguments and even something like `INT_MAX >> 2` (`>> 3` should have less digits then `sizeof(int) * 2`) will consume lots of memory warning could be safely discarded. I am not opposed to fixing that though, additional byte is not going to significantly increase memory footprint or make `powerline.c` slow.
2. Without the previous fix there actually is no way you may use `snprintf` output: `snprintf` is supposed to return amount of bytes *it would’ve written should the buffer is large enough*. It does *not* tell “how many bytes it has written into the buffer”. What is worse, on some systems `snprintf` returns `-1` in place of “amount of bytes it would’ve written should the buffer is large enough” if buffer is not actually large enough: http://demin.ws/blog/english/2013/01/28/use-snprintf-on-different-platforms/. Despite the article being four years old AFAIK “Windows 7” part is still applicable. Though powerline.c does not support any platforms with such non-standard behaviour.

cryptomilk:
Those failed checks do not seem to be related to my changes.

gcc complains about snrpintf() with:
```
gcc -Wall -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong client/
powerline.c -o scripts/powerline             
client/powerline.c: In function ‘main’:                                                                                                                                             
client/powerline.c:125:36: warning: ‘__builtin___snprintf_chk’ output may be truncated before the last format character [-Wformat-truncation=]
  snprintf(num_args, NUM_ARGS_SIZE, "%x", argc - 1);                
                                    ^~~~                                                  
In file included from /usr/include/stdio.h:939:0,                                         
                 from client/powerline.c:4:                                                                                                                                         
/usr/include/bits/stdio2.h:64:10: note: ‘__builtin___snprintf_chk’ output between 2 and 9 bytes into a destination of size 8
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,                    
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                    
        __bos (__s), __fmt, __va_arg_pack ());             
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ZyX-I:
@cryptomilk I do not understand what your comment is about. I was responding to commit comment which implicitly states that `snprintf` may *not* terminate the string with NUL (“snprintf needs to null-terminate the string, make sure it *can* do it”). It is not true, `snprintf` *always* terminates the string, truncating its output if needed.

ZyX-I:
Merged in f35c401 (edited commit comments).

PR #1816, opened 2017-08-08
fix spotify applescript track length

arslan2012:
Because duration of current track is returned by milliseconds from spotify.
it should be divided by 1000 before being displaying.

### DIFF
--- a/powerline/segments/common/players.py
+++ b/powerline/segments/common/players.py
@@ -341,7 +341,7 @@ class SpotifyAppleScriptPlayerSegment(PlayerSegment):
 						set track_name to name of current track
 						set artist_name to artist of current track
 						set album_name to album of current track
-						set track_length to duration of current track
+						set track_length to round ((duration of current track) / 1000) rounding down
 						set now_playing to "" & player state & "{0}" & album_name & "{0}" & artist_name & "{0}" & track_name & "{0}" & track_length
 						return now_playing
 					else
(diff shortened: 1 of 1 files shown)


ZyX-I:
It is better to do this in Python, apple script is lengthy.

ZyX-I:
By the way, spotify does not really export elapsed time via apple script or it was simply missed by the original author?

arslan2012:
Spotify exports elapsed time by name "player position"

arslan2012:
Ok, i reverted the apple script and changed another line

ZyX-I:
@arslan2012 Can you add elapsed time (and check that addition works)?

arslan2012:
@ZyX-I  I added it but nothing is happening, my status bar is same as before, I don't really know what's wrong.
```
def get_player_status(self, pl):
	status_delimiter = '-~`/='
	ascript = '''
		tell application "System Events"
			set process_list to (name of every process)
		end tell
			if process_list contains "Spotify" then
			tell application "Spotify"
				if player state is playing or player state is paused then
					set track_name to name of current track
					set artist_name to artist of current track
					set album_name to album of current track
					set track_length to duration of current track
					set now_playing to "" & player state & "{0}" & album_name & "{0}" & artist_name & "{0}" & track_name & "{0}" & track_length & "{0}" & player position
					return now_playing
				else
					return player state
				end if
				end tell
		else
			return "stopped"
		end if
	'''.format(status_delimiter)
		spotify = asrun(pl, ascript)
	if not asrun:
		return None
		spotify_status = spotify.split(status_delimiter)
	state = _convert_state(spotify_status[0])
	if state == 'stop':
		return None
	return {
		'state': state,
		'album': spotify_status[1],
		'artist': spotify_status[2],
		'title': spotify_status[3],
		'total': _convert_seconds(int(spotify_status[4])/1000),
		'elapsed': _convert_seconds(int(float(spotify_status[5])))
	}
```

PR #1821, opened 2017-08-19
Remove hashtag character at 48

taitung:
Fix last_pipe_status: bad array subscript

### DIFF
--- a/powerline/bindings/bash/powerline.sh
+++ b/powerline/bindings/bash/powerline.sh
@@ -45,7 +45,7 @@ _powerline_status_wrapper() {
 	local last_exit_code=$? last_pipe_status=( "${PIPESTATUS[@]}" )
 
 	if ! _powerline_has_pipestatus \
-	   || test "${#last_pipe_status[@]}" -eq "0" \
+	   || test "${last_pipe_status[@]}" -eq "0" \
 	   || test "$last_exit_code" != "${last_pipe_status[-1]}" ; then
 		last_pipe_status=()
 	fi


ZyX-I:
This change is completely incorrect, what code there means is “check whether array is not empty” and that code is exactly what is needed in bash to check array length. If on some of earlier bash versions this does not work as intended, change code so that it will work to check for array not being empty, *not* to check whether array contains exactly one element which is equal to zero, giving unwanted error message should “contains exactly one element” check be false.

taitung:
Thanks for the clarification. I will remove the PR

ZyX-I:
@taitung I think that check may be replaced with `test -z "${last_pipe_status[*]}"`, though only due to a nature of the values in the array: should `last_pipe_status` be equal to `( "" )` such check will incorrectly treat array as empty, but it is valid since `last_pipe_status` is expected to contain only numbers.

ZyX-I:
By the way, which bash version does not allow `${#arr[@]}`, but has `$PIPESTATUS`?

ZyX-I:
And also does that bash version allow using negative array indexes to index from the end of the array?

taitung:
@ZyX-I my bash version 4.4.12(1)-release. I revert my local copy to keep that hashtag. Afterwards,  I changed a bit that line but just locally verified it and it probably works okay. However, the solution is not that elegant. My purpose is just to let Powerline works on machines.  
`test "$last_exit_code" != "${last_pipe_status[${#last_pipe_status[@]} -1]}`

ZyX-I:
@taitung Check `echo $BASH_VERSION`, not `bash --version`. I doubt very highly that bash developers *removed* negative indexes or anything, and negative indexes work on 4.3.48.

taitung:
@ZyX-I echoing $BASH_VERSION is pretty much the same with bash --version, which is **4.4.12(1)-release**

ZyX-I:
@taitung Then are you sure it is needed to replace negative index with a math expression which is supposed to yield negative index (and thus error) if array is empty and work just like the replaced index if it is not?

Can you post a trace (`set -x`) with a error in an unmodified code?

taitung:
@ZyX-I hope following info will be helpful
```
Last login: Tue Aug 22 14:42:54 on ttys000
-bash: last_pipe_status: bad array subscript
 taitung  ~  set -x ls
++ _powerline_status_wrapper _powerline_set_prompt
++ last_pipe_status=("${PIPESTATUS[@]}")
++ local last_exit_code=0 last_pipe_status
++ _powerline_has_pipestatus
++ return 0
++ test 1 -eq 0
-bash: last_pipe_status: bad array subscript
++ test 0 '!=' ''
++ last_pipe_status=()
++ _powerline_set_prompt 0 ''
++ local last_exit_code=0
++ shift
++ local last_pipe_status=
++ shift
+++ jobs -p
+++ wc -l
++ local 'jobnum=       0'
+++ _powerline_prompt aboveleft 0 '' 0
+++ /Users/taitung/Library/Python/2.7/bin/powerline shell aboveleft --width=270 -r.bash --last-exit-code=0 --last-pipe-status= --jobnum=0 --renderer-arg=client_id=19854
++ PS1='\[\] taitung \[\] \[\]~ \[\] \[\]'
++ test -n 1
+++ _powerline_local_prompt left -r.bash 0 '' 0 continuation
+++ /Users/taitung/Library/Python/2.7/bin/powerline shell left -r.bash --last-exit-code=0 --last-pipe-status= --jobnum=0 --renderer-arg=client_id=19854 --renderer-arg=local_theme=continuation
++ PS2='\[\]            \[\] \[\]'
++ test -n 1
+++ _powerline_local_prompt left '' 0 '' 0 select
+++ /Users/taitung/Library/Python/2.7/bin/powerline shell left --last-exit-code=0 --last-pipe-status= --jobnum=0 --renderer-arg=client_id=19854 --renderer-arg=local_theme=select
++ PS3=' Select variant  '
++ return 0
++ update_terminal_cwd
++ local url_path=
++ local i ch hexch LC_CTYPE=C LC_ALL=
++ (( i = 0 ))
++ (( i < 13 ))
++ ch=/
++ [[ / =~ [/._~A-Za-z0-9-] ]]
++ url_path+=/
++ (( ++i ))
++ (( i < 13 ))
++ ch=U
++ [[ U =~ [/._~A-Za-z0-9-] ]]
++ url_path+=U
++ (( ++i ))
++ (( i < 13 ))
++ ch=s
++ [[ s =~ [/._~A-Za-z0-9-] ]]
++ url_path+=s
++ (( ++i ))
++ (( i < 13 ))
++ ch=e
++ [[ e =~ [/._~A-Za-z0-9-] ]]
++ url_path+=e
++ (( ++i ))
++ (( i < 13 ))
++ ch=r
++ [[ r =~ [/._~A-Za-z0-9-] ]]
++ url_path+=r
++ (( ++i ))
++ (( i < 13 ))
++ ch=s
++ [[ s =~ [/._~A-Za-z0-9-] ]]
++ url_path+=s
++ (( ++i ))
++ (( i < 13 ))
++ ch=/
++ [[ / =~ [/._~A-Za-z0-9-] ]]
++ url_path+=/
++ (( ++i ))
++ (( i < 13 ))
++ ch=g
++ [[ g =~ [/._~A-Za-z0-9-] ]]
++ url_path+=g
++ (( ++i ))
++ (( i < 13 ))
++ ch=c
++ [[ c =~ [/._~A-Za-z0-9-] ]]
++ url_path+=c
++ (( ++i ))
++ (( i < 13 ))
++ ch=h
++ [[ h =~ [/._~A-Za-z0-9-] ]]
++ url_path+=h
++ (( ++i ))
++ (( i < 13 ))
++ ch=i
++ [[ i =~ [/._~A-Za-z0-9-] ]]
++ url_path+=i
++ (( ++i ))
++ (( i < 13 ))
++ ch=o
++ [[ o =~ [/._~A-Za-z0-9-] ]]
++ url_path+=o
++ (( ++i ))
++ (( i < 13 ))
++ ch=u
++ [[ u =~ [/._~A-Za-z0-9-] ]]
++ url_path+=u
++ (( ++i ))
++ (( i < 13 ))
++ printf '\e]7;%s\a' file://MacBook.local/Users/taitung
```

taitung:
@ZyX-I my apology. I thought I was doing it correctly about the `echo $BASH_VERSION`. I googled a bit and found one article here https://apple.stackexchange.com/questions/193411/update-bash-to-version-4-0-on-osx about newer bash version allowed on macOS. I changed to zsh after facing issue of bad subscript. So, it is just I change to bash and then echo $BASH_VERSION. So, the version number is almost identical. However, that is not correct. Following that article, on my macOS it works fine now with bash as my default shell.

ZyX-I:
I see in the changelog that negative indexes were introduced in 4.3, so I would need to get rid of it anyway. Not sure though whether there is (no) bash version which has PIPESTATUS, but does not support math in subscripts, so it would probably be better to use

```shell
test "$last_exit_code" != "${last_pipe_status[$(( ${#last_pipe_status[@]} - 1 ))]}
```

taitung:
go it. thank you @ZyX-I

PR #1827, opened 2017-09-02
fix data type error

storyn26383:


ZyX-I:
It is supposed to be `list` or `set` in this case, not `array`.

PR #1830, opened 2017-09-11
Fix typo

i5ar:


s-ol:
That's actually not a typo, a lot of programd have flags that work with omitted spaces like this.
Did you get an error with this?

I just tested it on my phone with `python3 -mhttp.server` and it worked.

i5ar:
You're right, it didn't work for me for a different reason.

PR #1851, opened 2017-10-23
Added Vim segment plugin for the Asynchronous Linter Engine (ALE)

pmacosta:
This segment has the same functionality and interface as the Syntastic
Vim segment plugin, and the code is largely adapted from it

### DIFF
--- a/powerline/segments/vim/plugin/ale.py
+++ b/powerline/segments/vim/plugin/ale.py
@@ -1,3 +1,4 @@
+# vim:fileencoding=utf-8:noet
 from __future__ import (unicode_literals, division, absolute_import, print_function)
 
 try:
@@ -23,14 +24,14 @@ def ale(segment_info, pl, err_format='ERR: ln {first_line} ({num}) ', warn_forma
 	'''
 	if not (vim_global_exists('ale_enabled') and int(vim.eval('g:ale_enabled'))):
 		return None
-	has_errors = int(vim.eval('ale#statusline#Count('+str(segment_info['bufnr'])+').total'))
+	has_errors = int(vim.eval('ale#statusline#Count(' + str(segment_info['bufnr']) + ').total'))
 	if not has_errors:
 		return
 	error = None
 	warning = None
 	errors_count = 0
 	warnings_count = 0
-	for issue in vim.eval('ale#engine#GetLoclist('+str(segment_info['bufnr'])+')'):
+	for issue in vim.eval('ale#engine#GetLoclist(' + str(segment_info['bufnr']) + ')'):
 		if issue['type'] == 'E':
 			error = error or issue
 			errors_count += 1
--- a/powerline/segments/vim/plugin/ale.py
+++ b/powerline/segments/vim/plugin/ale.py
@@ -1,42 +1,17 @@
-# vim:fileencoding=utf-8:noet
-"""
-Asynchronous Lint Engine (ALE) plugin
+from __future__ import (unicode_literals, division, absolute_import, print_function)
 
-Largely adapted from Sytastic plugin
-"""
-
-from __future__ import (
-	unicode_literals, division, absolute_import, print_function
-)
 try:
 	import vim
 except ImportError:
 	vim = object()
+
 from powerline.bindings.vim import vim_global_exists
 from powerline.theme import requires_segment_info
 
 
-# To turn on logging, in the Powerline configuration fie config.json,
-# in the "common" key, add log_file and log_level key-value pairs as
-# appropriate and desired, for example:
-#
-# "common": {
-#	  "term_truecolor": false,
-#	  "log_file": "/mylogdir/powerline.log",
-#	  "log_level": "INFO"
-# },
-#
-# and then turn on debug internal variable to True
 @requires_segment_info
-def ale(
-	segment_info,
-	pl,
-	err_format='ERR:  {first_line} ({num}) ',
-	warn_format='WARN:  {first_line} ({num}) '
-):
-	"""
-	Show whether Asynchronous Lint Engine (ALE) has found any errors
-	or warnings
+def ale(segment_info, pl, err_format='ERR: ln {first_line} ({num}) ', warn_format='WARN: ln {first_line} ({num}) '):
+	'''Show whether ALE has found any errors or warnings
 
 	:param str err_format:
 		Format string for errors.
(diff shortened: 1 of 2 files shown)


powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+	:param str err_format:
+		Format string for errors.
+
+	:param str warn_format:
+		Format string for warnings.
+
+	Highlight groups used: ``ale:warning`` or
+	``warning``, ``ale:error`` or ``error``.
+	"""
+	# pylint: disable=C0103,W0613
```
ZyX-I: Do not add `pylint` comments, powerline is not using it.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+	debug = False
+	bufnr = str(segment_info['bufnr'])
+	ale_enabled_exists = vim_global_exists('ale_enabled')
+	ale_enabled = ale_enabled_exists and int(vim.eval('g:ale_enabled'))
+	if debug:
```
ZyX-I: This should not be hidden under `if debug`, you should use `pl.debug` instead.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+		pl.info('ale_enabled_exists: {0}'.format(ale_enabled_exists))
```
ZyX-I: Do not use `.format`, `PowerlineLogger` does already use it internally.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+#	  "log_level": "INFO"
+# },
+#
+# and then turn on debug internal variable to True
+@requires_segment_info
+def ale(
+	segment_info,
+	pl,
+	err_format='ERR:  {first_line} ({num}) ',
+	warn_format='WARN:  {first_line} ({num}) '
```
ZyX-I: Should not use non-ASCII here any longer, actually I need to remove that from syntastic as well. All non-ASCII should go to relevant top-level theme files like `/powerline/config_files/themes/powerline.json`.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+	has_errors = int(vim.eval(cmd))
+	if debug:
+		pl.info('has_errors: {0}'.format(has_errors))
+	if not has_errors:
+		return
+	cmd = 'ale#engine#GetLoclist({0})'.format(bufnr)
+	if debug:
+		pl.info('cmd: {0}'.format(cmd))
+	issues = vim.eval(cmd)
+	errors, warnings = [], []
```
ZyX-I: Do not have multiple definitions on one line unless you have to (i.e. when unpacked tuple returned by the function). Also do not use lists, you need *one* entry of the whole list and list length after all, you don’t need the entire list.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+	or warnings
+
+	:param str err_format:
+		Format string for errors.
+
+	:param str warn_format:
+		Format string for warnings.
+
+	Highlight groups used: ``ale:warning`` or
+	``warning``, ``ale:error`` or ``error``.
```
ZyX-I: Must not wrap, this comment is here for the *linter* so the format needs to be *exact*, like described in develop/segments.rst.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
...
+# and then turn on debug internal variable to True
+@requires_segment_info
+def ale(
+	segment_info,
+	pl,
+	err_format='ERR:  {first_line} ({num}) ',
+	warn_format='WARN:  {first_line} ({num}) '
+):
+	"""
+	Show whether Asynchronous Lint Engine (ALE) has found any errors
```
ZyX-I: Too long summary, summary not on the same line with quotes, using double quotes in place of single.

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,97 @@
+# vim:fileencoding=utf-8:noet
+"""
+Asynchronous Lint Engine (ALE) plugin
+
+Largely adapted from Sytastic plugin
+"""
+
+from __future__ import (
+	unicode_literals, division, absolute_import, print_function
+)
```
ZyX-I: Just copy the header from syntastic: no wrapping, more empty lines and no docstring.

pmacosta: Fixed in new commit

powerline/segments/vim/plugin/ale.py
pmacosta: Fixed in new commit, file is ASCII-encoded now

powerline/segments/vim/plugin/ale.py
pmacosta: Fixed in new commit

powerline/segments/vim/plugin/ale.py
pmacosta: Fixed in new commit

powerline/segments/vim/plugin/ale.py
pmacosta: Fixed in new commit

powerline/segments/vim/plugin/ale.py
pmacosta: Actually removed all debug logging, in hindsight this was only needed to get the ALE interfacing correct, and it only clutters code now. Removal also makes Syntastic and ALE plugin files more similar.

powerline/segments/vim/plugin/ale.py
pmacosta: Location list is traversed once, first error and warning are captured and from then on only error and warning count is kept

powerline/segments/vim/plugin/ale.py
pmacosta: Actually removed all debug logging, in hindsight this was only needed to get the ALE interfacing correct, and it only clutters code now. Removal also makes Syntastic and ALE plugin files more similar.

powerline/segments/vim/plugin/ale.py
pmacosta: Removed all logging code

powerline/segments/vim/plugin/ale.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import (unicode_literals, division, absolute_import, print_function)
```
ZyX-I: Now it is missing modeline. LGTM, except for this and lack of spaces around some operators (`+`) issues.

pmacosta: Sorry, brought back modeline and added spaces around plus signs

pmacosta: Happy to squash the commits if that is your preference

PR #1853, opened 2017-10-31
Corrected interface lookups to work with systemd

hrabbit:
This should allow lookups to be performed against the new systemd interface (as well as all other interfaces) simply by comparing the prefix of the interface name rather than the first characters only. This also allows for enx interfaces that use the mac address as the identifier and throws the modifier for a spin by returning a crazy long integer.

Reference: https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L20

### DIFF
--- a/powerline/segments/common/net.py
+++ b/powerline/segments/common/net.py
@@ -85,21 +85,16 @@ else:
 		'vboxnet':  -5,  # VirtualBox bridge interface       : vboxnet0
 	}
 
-	_interface_start_re = re.compile(r'^([a-z]+?)(\d|$)')
-
 	def _interface_key(interface):
-		match = _interface_start_re.match(interface)
-		if match:
-			try:
-				base = _interface_starts[match.group(1)] * 100
-			except KeyError:
-				base = 500
-			if match.group(2):
-				return base - int(match.group(2))
-			else:
-				return base
-		else:
-			return 0
+		for prefix,score in _interface_starts.items():
+			if(interface.startswith(prefix)):
+				try:
+					# This prevents a huge negative when the interface name is a mac address
+					modifier = int(re.sub('[^0-9]+', '', interface)) < 10 or 0
+				except:
+					modifier = 0
+				return score * 100 - modifier
+		return 0
 
 	def internal_ip(pl, interface='auto', ipv=4):
 		family = netifaces.AF_INET6 if ipv == 6 else netifaces.AF_INET


powerline/segments/common/net.py
```diff
@@ -85,21 +85,16 @@ def internal_ip(pl, interface='auto', ipv=4):
...
-			else:
-				return base
-		else:
-			return 0
+		for prefix,score in _interface_starts.items():
+			if(interface.startswith(prefix)):
+				try:
+					# This prevents a huge negative when the interface name is a mac address
+					modifier = int(re.sub('[^0-9]+', '', interface)) < 10 or 0
+				except:
```
ZyX-I: Do not use `re.sub`, or any other method from `re` which is not `re.compile`. Also catch specific exceptions.

powerline/segments/common/net.py
```diff
@@ -85,21 +85,16 @@ def internal_ip(pl, interface='auto', ipv=4):
...
-			try:
-				base = _interface_starts[match.group(1)] * 100
-			except KeyError:
-				base = 500
-			if match.group(2):
-				return base - int(match.group(2))
-			else:
-				return base
-		else:
-			return 0
```
ZyX-I: And need an update to tests (tests/test_python/test_segments.py, class TestNet) with new interface examples.

powerline/segments/common/net.py
```diff
@@ -85,21 +85,16 @@ def internal_ip(pl, interface='auto', ipv=4):
...
+		for prefix,score in _interface_starts.items():
+			if(interface.startswith(prefix)):
```
ZyX-I: Do not use `if()`, it needs no parenthesis.

powerline/segments/common/net.py
```diff
@@ -85,21 +85,16 @@ def internal_ip(pl, interface='auto', ipv=4):
...
-				base = _interface_starts[match.group(1)] * 100
-			except KeyError:
-				base = 500
-			if match.group(2):
-				return base - int(match.group(2))
-			else:
-				return base
-		else:
-			return 0
+		for prefix,score in _interface_starts.items():
```
ZyX-I: Missing space.

powerline/segments/common/net.py
ZyX-I: (Specifically, the ones which make the current code fail; I do not think that your variant is the best one.)

powerline/segments/common/net.py
hrabbit: ZyX-I, thanks for the feedback. To be honest, I didn't expect it to end in a humans lap after the Travis failures. (Not a Python developer in the slightest)

I'll clean this up and see about putting in another pull again soon.

PR #1867, opened 2017-12-21
tell people how to enable Vim statusline

xinbenlv:
I bet author got this question a lot, why just tell people how to explain it?


ZyX-I:
`'laststatus'` is a user setting, having it have particular value is not required to use powerline, that only controls when it is seen, so I don’t want to see it directly in any code blocks which tell how to enable powerline: documentation there is closer to technical documentation then to a user guide in any case. There are also another issues:

1. The code block you just edited is not a single variant of enabling powerline.
2. `#` is not a comment character in VimL.

And note stands out more then code block if somebody actually reads documentation.

Issue #1863, opened 2017-12-02
Powerline ignores the term_truecolor setting for tmux statusline


kmoschcau:
Hi,
I have an odd little case where powerline seems to ignore the "term_truecolor" global setting just for the tmux statusline. This happens even though my terminal and tmux are perfectly capable of using truecolor.
I have configured powerline to use the gruvbox theme (https://github.com/morhetz/gruvbox) for everything I use. It works everywhere except the powerline tmux statusline. It's probably best to show some screenshots.

This shows two neovim instances inside tmux. The colors for neovim are correct, the ones for the statusline are not. The outermost elements in the statusline should have the same color as the mode element in neovim.
![selection_001](https://user-images.githubusercontent.com/9497302/33515235-95b0d108-d760-11e7-85ac-151ecb9c5b32.png)
(Disclaimer: I did not get powerline to work with neovim. Neovim just uses the plain gruvbox theme and vim-airline. It uses the same colors though. I just included this as a proof, that truecolor works inside tmux.)

This shows the same tmux instance with a terminal. As you can see again, the terminal PS1 uses the right colors.
![selection_002](https://user-images.githubusercontent.com/9497302/33515263-21ef013a-d761-11e7-9d4c-2e3ffd279a2c.png)

Here now is the main reason why I think the problem is with powerline: Before I source the powerline binding for tmux, I set up a sort of fallback in my .tmux.conf with the right colors. And look, they actually work!
![selection_003](https://user-images.githubusercontent.com/9497302/33515295-7493f224-d761-11e7-9c9e-52c369c395c4.png)

And now what makes me think that powerline ignores the "term_truecolor" setting for tmux: This is how my PS1 looks, when I set "term_truecolor" to false.
![selection_004](https://user-images.githubusercontent.com/9497302/33515323-c65c61fe-d761-11e7-91cb-ee5e00001915.png)
That looks pretty much the same to me.

The gruvbox theme provides colors for truecolor and 256 colors. I set both of them in my colors configuration for powerline. Here is my colors.json:
```json
{
  "colors": {
    "gruvbox_dark:0_bg":            [235, "282828"],
    "gruvbox_dark:8_gray":          [245, "928374"],

    "gruvbox_dark:1_dark_red":      [124, "cc241d"],
    "gruvbox_dark:9_light_red":     [167, "fb4934"],

    "gruvbox_dark:2_dark_green":    [106, "98971a"],
    "gruvbox_dark:10_light_green":  [142, "b8bb26"],

[...]

PR #1869, opened 2017-12-29
Tmux truecolor support

Genora51:
Fixes issue #1863.

This PR adds some code to `powerline/renderers/tmux.py` that checks if truecolor is enabled, and if so, uses the truecolor 24-bit hex colour rather than a 256-colour version.


Genora51:
It seems to not be working correctly - will look into it but until then I am closing the PR.

PR #1870, opened 2018-01-10
(fix) updated linux spotify app dbus config

gffranco:
Updating linux dbus config according to Spotify client version: 1.0.69.336.g7edcc575.

Now player segment works as expected.

### DIFF
--- a/powerline/segments/common/players.py
+++ b/powerline/segments/common/players.py
@@ -311,10 +311,10 @@ class SpotifyDbusPlayerSegment(PlayerSegment):
 		return _get_dbus_player_status(
 			pl=pl,
 			player_name='Spotify',
-			bus_name='com.spotify.qt',
-			player_path='/',
+			bus_name='org.mpris.MediaPlayer2.spotify',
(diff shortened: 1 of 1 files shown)


ZyX-I:
That needs to do version detection.

gffranco:
Hello @ZyX-I and thanks for looking into this. 

From my understanding, there is no way for us to check the client version from powerline and I'm also not sure what version changed the dbus configuration, which makes version checking this not reliable.

What I had in mind to improve this is, since we know these 2 dbus sessions configurations used by Spotify is to check whether the 'com.spotify.qt' or 'org.mpris.MediaPlayer2.spotify' session bus is present and use the first found.

Do you think this would be good enough?

Another possibility is to simply document this and point people to use the Generic Dbus segment with the correct arguments.

ZyX-I:
Yes, if you can try new session bus and old one on failure rather then explicit version check it would be better and more reliable, but version check seemed simpler, there are some functions for getting shell command output which should work unless spotify is being strange and does something other then just outputting version information when called using something like `spotify --version`.

Issue #1754, opened 2017-04-20
Spotify segment don't work


victordev:
log:
`2017-04-19 21:27:08,988:DEBUG:tmux:watcher:Trying to use inotify watcher`
`2017-04-19 21:27:08,988:DEBUG:tmux:watcher:Trying to use inotify watcher`

powerline-lint:
```
Script started on Qua 19 Abr 2017 21:34:06 BRT

Error while checking theme (key /segments/right/list item 1/function)
found highlight groups list (player_fallback, player) with all groups not defined in some colorschemes
(Group names were taken from function documentation.)
  in "/home/victor/.config/powerline/themes/tmux/default.json", line 12, column 56:
     ... werline.segments.common.players.spotify_dbus",
                                         ^

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player_fallback not defined in the following colorschemes: default, solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player not defined in the following colorschemes: solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight groups list (player_play, player) with all groups not defined in some colorschemes
(Group names were taken from function documentation.)
  in "/home/victor/.config/powerline/themes/tmux/default.json", line 12, column 56:
     ... werline.segments.common.players.spotify_dbus",
                                         ^

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player_play not defined in the following colorschemes: default, solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player not defined in the following colorschemes: solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight groups list (player_pause, player) with all groups not defined in some colorschemes
(Group names were taken from function documentation.)
  in "/home/victor/.config/powerline/themes/tmux/default.json", line 12, column 56:
     ... werline.segments.common.players.spotify_dbus",
                                         ^

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player_pause not defined in the following colorschemes: default, solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight group player not defined in the following colorschemes: solarized

Error while checking theme (key /segments/right/list item 1/function)
found highlight groups list (player_stop, player) with all groups not defined in some colorschemes
(Group names were taken from function documentation.)
  in "/home/victor/.config/powerline/themes/tmux/default.json", line 12, column 56:
     ... werline.segments.common.players.spotify_dbus",
                                         ^

[...]

PR #1893, opened 2018-03-14
(fix) updated spotify segment to support both new and old protocol

StopMotionCuber:
This PR is based off #1870 and adressing Issue #1754


ZyX-I:
It should try new protocol first and not old.

StopMotionCuber:
OK, have changed that.
I thought that I wouldn't break functionality if I checked for the old one first so that is why I chose that approach (+ I don't really think it matters as nobody has 2 instances of spotify on the same machine with a different version running)

Issue #1895, opened 2018-03-15
Tmux client Environment


StopMotionCuber:
I got a question:

Is it possible to get the tmux client environment (the environment that you got when you start the tmux session or attach to it)?

I'm currently using the player segment and would like to add a feature to disable it depending on whether the client is run via SSH (because I do not want to know whether my remote machine is playing music).
If this is possible but needs further programming I'd be open to do this work. In this case I would be happy if I could get some advise on what files to look at.

ZyX-I:
`tmux` AFAIK does not save environment it starts with side from the very first run: when server is started it does inherit parent environment. All other `tmux` runs it merely directs tmux server what to do, then tmux *server* forks, obviously having child gain server’s environment (not sure what stopped that other tmux runs from just dumping all environment on server to be used in child later: this works fine with even powerline-daemon which ought to run on each prompt). You *can* write a plugin for the *shell* run inside a client which will notify about environment changes, powerline actually does that for current working directory, but this approach is far from universal and trying to use the existing code for all of the environment present in shell is going to bite performance hard and create lots of environment variables in tmux which is not healthy.

StopMotionCuber:
I found a way that can detect whether a tmux-session is running via SSH:
`tmux list-clients -F '{client_pid}` lists you all the clients process IDs.
You can go upwards in the process tree and search for the ssh daemon.
I did not find a way though to identify within powerline for which tmux-client it is rendering.

Also I wanted to reference #1750 as it describes the same core problem

ZyX-I:
Based on what I know and what is working right now you have two options:

1. See how pane_id render option and width option are defined, do the same thing.
2. Newer tmux versions: <code>\`tmux display -p \\"#\\"\\"…\\"\`</code> blocks should be replaceable with just `\"#…\"` right in the command, the whole quotes mess is there primary to avoid tmux expanding `#…` in newer versions while also making code continue to work in older versions. There is mechanizm which allows having to define different options for different tmux versions, but using it for such kind of problems makes maintaining bindings harder.
3. Calls to `tmux` right in powerline code should also work, as long as you supply correct environment (distributed via `segment_info`, `os.environ` is not being updated in powerline-daemon).

Issue #858, opened 2014-04-09
tmux hostname segment and ssh.


kierun:
If I have the `"only_if_ssh"=true` option set for the hostname segment in the default tmux configuration, then I expect to see the hostname only if I am using ssh.  This is work as expected provided that there is no other tmux session running on the host. If I have a session already running, then the hostname does not show.

Steps:
- Log into your machine via the console.
- Starts tmux: hostnane does not show, as expected.
- Log into your machine via ssh.
- Start tmux: hostname does not show whereas was expecting it to show.

Bug?  Or am I doing something weird.
##

StayPirate:
`SSH_CLIENT` got deprecated in favor of `SSH_CONNECTION`. The latter is available even within tmux, IMHO a small change from

https://github.com/powerline/powerline/blob/8b07f6396127f50fb73de9020912ae252d3ed170/powerline/segments/common/net.py#L30

to

```python
 if only_if_ssh and not segment_info['environ'].get('SSH_CONNECTION'): 
```

might solve the problem.

PR #1897, opened 2018-03-23
Added Tmux Selector

StopMotionCuber:
Implementation will show segments depending on whether the current active tmux client is connected via SSH
This still has the flaw that it does not draw independent status bars to the screens, but draws the same depending on the status of the last active pane (whether that has is connected via SSH)
References #1895 and #858


powerline/selectors/tmux.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def tmux_via_ssh(pl, segment_info, mode, caching=True, default=False):
+	'''Returns True if tmux client is connected via SSH
+	'''
+	pid = get_tmux_output(pl, "display", "-p", "#{client_pid}")
+	if not pid.isnumeric():
+		return default
+	if caching:
+		if pid in _cached_clients:
```
ZyX-I: This logical to write using EAFP principle: just return `_cached_clients[pid]` and if that throws `KeyError` proceed with using `psutil`.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,32 @@
...
+			return _cached_clients[pid]
+	try:
+		pstree_out = subprocess.check_output(("pstree", "-s", pid)).decode()
```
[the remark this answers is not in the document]
ZyX-I: Actually don’t use `pstree`, as well as any regular expressions. `psutil.Process` class provides everything needed, it works in-process, it should be more portable and it would not query all the parents, just up to the one that happens to match `sshd`. Just need to mention requirements in the documentation. `pstree` is the requirement as well though.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,32 @@
...
+_cached_clients = {}
+# This Regex parses pstree -s <Process_ID> information
+# It will match, if any process name is sshd
+SSHD_REGEX = re.compile(r"(?:^|\W)sshd(?:^|\W)")
+
+
+def tmux_via_ssh(pl, segment_info, mode, caching=True, default=False):
+	'''Returns True if tmux client is connected via SSH
+	'''
+	pid = get_tmux_output(pl, "display", "-p", "#{client_pid}")
```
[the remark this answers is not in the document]
StopMotionCuber: Is there a better way of doing it?
For me this worked okay-ish (it got the `display -p #{client_pid}` output for the current active tmux window so that it would not show excluded segments when I was connected via SSH to the machine and working actively on it), also when using the powerline-daemon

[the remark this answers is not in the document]
ZyX-I: You need another tmux instance *not* run from ssh for that to not be OK. This does not mean that variant is acceptable, you would need to refactor relevant code.

StopMotionCuber:
I have done changes so that it works now consistently.
I yet have to add backwards compatibility for tmux < 1.7 and add documentation (including psutil requirement).
For my machine this is working in a nice way. Tell me if you have any concerns about this code

StopMotionCuber:
I added documentation now for the tmux SSH selector.
This works like a charm for my tmux version and it should work for every tmux-version >= 2.1.
@ZyX-I I would really appreciate if you could review these changes.

powerline/commands/main.py
```diff
@@ -107,6 +107,10 @@ def get_argparser(ArgumentParser=argparse.ArgumentParser):
 		'-w', '--width', type=int,
 		help='Maximum prompt with. Triggers truncation of some segments.'
 	)
+	parser.add_argument(
+		'--process-id', metavar='PID', type=int,
+		help='process_ID for the desired process'
```
ZyX-I: Documentation should start with a capital, end with dot, not use identifiers where they are not meant to (what’s wrong with “Process ID”?) and be more descriptive then “desired process”. I have no idea what “desired process” is supposed to mean.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+_cached_clients = {}
+
+def tmux_via_ssh(pl, segment_info, mode):
+	'''Returns True if tmux client is connected via SSH
+	
+	Requires the ``psutil`` module.
+	
+	Works only for tmux versions >= 2.1'''
```
ZyX-I: This is tied to `--process-id` being supplied on command-line, not to tmux version in any fashion. Also closing `'''` needs to be on its own line and it is not qualified to be a sentence end.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
+# vim:fileencoding=utf-8:noet
```
ZyX-I: And I would actually suggest moving that to `powerline/selectors/proc.py` and get rid of any references to tmux here. Neither idea of the process tree traversal nor current implementation of the selector is really tmux-specific.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+from powerline.bindings.tmux import get_tmux_output
```
ZyX-I: Unused import.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+_cached_clients = {}
```
ZyX-I: Keys are process identifiers.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+
+def tmux_via_ssh(pl, segment_info, mode):
```
ZyX-I: `in_ssh`.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+	'''Returns True if tmux client is connected via SSH
```
ZyX-I: … if ``--process-id`` has sshd as its ancestor

powerline/selectors/tmux.py
StopMotionCuber: #{client_pid} got introduced with tmux 2.1, so it is not supported before

ZyX-I: This is tmux bindings implementation detail. All what *selector* cares about is whether `--process-id` is supplied or not, if it is it would be completely happy without even having tmux installed.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+	
+	Requires the ``psutil`` module.
+	
+	Works only for tmux versions >= 2.1'''
+	try:
+		import psutil
+	except ImportError:
+		pl.warn('Module "psutil" is not installed, thus we cannot detect whether tmux is run via ssh')
```
ZyX-I: This check should be performed at module start. See `cpu_load_percent` segment in powerline/segments/common/sys: depending on whether import is successfull it either creates an actual class or a dummy which does nothing, but a warning with `None` return.

powerline/selectors/tmux.py
```diff
@@ -0,0 +1,34 @@
...
+		return False
+	process_id = segment_info['args'].process_id
+	if process_id is None:
+		pl.warn("Process ID was not passed, maybe you are using tmux version < 2.1")
```
ZyX-I: Remove “, maybe …”.

powerline/commands/main.py
StopMotionCuber: Would 'The process ID of process that renders powerline.' be okay for this?

powerline/selectors/tmux.py
StopMotionCuber: The idea is good, however this requires the exclude function to be named `powerline.selectors.proc.is_ssh`.
I would propose to change this into a file named common.py and add this module to the general search path for the selectors

StopMotionCuber:
Thanks for your Feedback @ZyX-I.
I changed my code according to your requests and did some refactoring to have a common selector module.

ZyX-I:
Do not add second default selectors module. There must not be two default modules out there. Also if you were to do that you also need to

- Add tests for new behaviour.
- Change linter.
- Change documentation.

StopMotionCuber:
The problem with the current approach is that the default module is depending on your current powerline renderer.
For that and my usecase I would have to add it back to `tmux.py` which would be not portable for other renderers (or they had to add `powerline.selectors.tmux.is_ssh` in their config. This is too confusing as  tmux is not even used).
The other approach would be to always be explicit with `powerline.selectors.common.is_ssh` as include/exclude_function. I do think this is intuitive for a configuration file either.
If you have a suggestion for a better approach then hit me up, otherwise I will have a look at further needed changes.

ZyX-I:
This is not the first place where path to function is (or needs to be) full with module, and it will not be the last. But AFAIR there are exactly no places where there are two default modules. I would happily remove *all* default modules completely everywhere, this makes functions easier to discover when needed and configuration file is not a thing that is edited too frequently to bother, but this is backwards incompatible and also not all people share same opinion.

PR #1907, opened 2018-05-08
Transparency doesn't need to be disabled

yparisien:
I seems that disable transparency is no more mandatory to keep good colors.

### DIFF
--- a/docs/source/troubleshooting/osx.rst
+++ b/docs/source/troubleshooting/osx.rst
@@ -26,8 +26,8 @@ The colors look weird in iTerm2!
 --------------------------------
 
 * The arrows may have the wrong colors if you have changed the “minimum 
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting/osx.rst
```diff
@@ -27,7 +27,7 @@ The colors look weird in iTerm2!
 
 * The arrows may have the wrong colors if you have changed the “minimum 
   contrast” slider in the color tab of your OS X settings.
-* Please disable background transparency to resolve this issue.
+* If you're using transparency, check "Keep background colors opaque"
```
ZyX-I: Why did you remove dot? Also note what is used as quotes just the line above and as ampersand on line 8.

PR #1915, opened 2018-06-05
Recognize Terminal-Job mode (added in vim 8)

dkrieger:
Closes https://github.com/powerline/powerline/issues/1916


dkrieger:
I amended the commit to address your concerns

PR #1938, opened 2018-08-20
Harden random renderer-arg in powerline-setup.fish

simonfxr:
Select the random renderer-arg in _powerline_update from a bigger domain. The previous version was overly optimistic, by ignoring the birthday paradox, so the actual collision probability could have been much higher. This is a very small fix, so it should not break anything :-)


ZyX-I:
It should actually be `%self` there (preferably without any subshells if possible: `'foo'%self` does not work), like in bash bindings which have `$$` there: that should have exactly zero probability of having two shells on one machine with same client ID in most common setups (I do know that there are PID namespaces in linux, but normally you would not have separate PID namespace and still share access to the location of socket: feature is used primary for containers). Not sure why I did not use that earlier, perhaps this was introduced in later versions? (Or, at least, was error message pointing to `%self` when trying to use `$$`.)

simonfxr:
Okay, this is indeed even simpler and more elegant! The original comment in the code just got me thinking.

PR #1945, opened 2018-09-20
doc: Correct links to vim-airline

chrisbra:


chrisbra:
as this is only a doc-fix I don't really understand the breakage of the CI

ZyX-I:
It looks like some travis issue: VM just hang up for ten minutes before even starting to execute .travis.yml.

PR #1956, opened 2018-10-20
Update docs

ArtFlag:
Hello, 

As a new user of Powerline, which I find great, I had issues with the install docs and I had to use other online tutorials to set it up.

I'm opening a first PR to submit readability updates to the installation page and the OSX install/bash sections. 
I would be happy to send in deeper updates to the docs if you find it useful, and potentially improve the navigation of the docs.

For the review, I suggest checking the output directly, as proof reading docs on Github is pretty tedious when blocks of text have been moved around.


ArtFlag:
Sorry about that, should be better now.

PR #1991, opened 2019-03-27
Support for tmux version "next-2.9"

GHPS:


GHPS:
While the code in the commit fixes the current problem it perhaps would be reasonable
to replace the whole parsing section with a more robust solution like

`prefix,major,minor,postfix=re.findall(r'(\D*)(\d*)\.(\d*)(\D*)',stVersion)[0] `

This piece of code works with all kinds of version numbers, one or two digits,
prefix and postfix parts.

alealv:
Thanks. It worked for me

orcinus:
Why was this never merged?
The problem is still present in 2023.

GHPS:
> Why was this never merged?

Thanks for the support - looking at my own four year old snippet of
simple clean code, the second PR addressing the same issue and the
latest bug report, it is hard to explain why this hasn't been merged.

@PH111P please merge any of the PRs and close the issue...

PH111P:
This never got merged because a different fix got merged four years ago…

https://github.com/powerline/powerline/pull/2033

I guess it is indeed time to close this PR, though.

Issue #1999, opened 2019-04-11
Doesn't work in NeoVim


RossComputerPerson:
When installed from pacman on Arch Linux, I get this error when opening neovim:
```
Traceback (most recent call last):
  File "<string>", line 4, in <module>
  File "/usr/lib/python3.7/site-packages/powerline/vim.py", line 15, in <module>
    from powerline.bindings.vim import vim_get_func, vim_getvar, get_vim_encoding, python_to_vim
  File "/usr/lib/python3.7/site-packages/powerline/bindings/vim/__init__.py", line 44, in <module>
    vim_encoding = get_vim_encoding()
  File "/usr/lib/python3.7/site-packages/powerline/bindings/vim/__init__.py", line 25, in get_vim_encoding
    return vim.options['encoding'].decode('ascii') or 'ascii'
AttributeError: 'str' object has no attribute 'decode'
An error occurred while importing powerline module.
This could be caused by invalid sys.path setting,
or by an incompatible Python version (powerline requires
Python 2.6, 2.7 or 3.2 and later to work). Please consult
the troubleshooting section in the documentation for
possible solutions.
Error detected while processing function provider#python3#Call:
line   18:
Traceback (most recent call last):
  File "<string>", line 53, in <module>
  File "<string>", line 40, in powerline_troubleshoot
  File "/usr/lib/python3.7/site-packages/powerline/lint/__init__.py", line 12, in <module>
    from powerline.segments.vim import vim_modes
  File "/usr/lib/python3.7/site-packages/powerline/segments/vim/__init__.py", line 16, in <module>
    from powerline.bindings.vim import (vim_get_func, getbufvar, vim_getbufoption,
  File "/usr/lib/python3.7/site-packages/powerline/bindings/vim/__init__.py", line 44, in <module>
    vim_encoding = get_vim_encoding()
  File "/usr/lib/python3.7/site-packages/powerline/bindings/vim/__init__.py", line 25, in get_vim_encoding
    return vim.options['encoding'].decode('ascii') or 'ascii'
AttributeError: 'str' object has no attribute 'decode'
```
Everything is up to date.

PH111P:
See #1287

Issue #1287, opened 2015-01-27
Add support for NeoVim


abstiles:
I understand that this may not be immediately feasible due to the still-unstable nature of NeoVim, but I'd like to make the request anyway.

First, powerline fails when it tries to access the bindeval method which NeoVim does not support. I attempted to do some initial debugging, and discovered that Powerline relies on Vim's 7.4 Python API when it detects a version greater than 703. NeoVim reports as version 704 right now, but has not added support for Vim's 7.4 Python API (see discussion on neovim/neovim#1898), which apparently includes the bindeval method. After attempting to work around this by modifying powerline's version checking, I'm still seeing the following issue:

```
Error detected while processing function PowerlinePyeval..provider#python#Call:
line    1:
NameError("name 'powerline' is not defined",)
```

I'm not sure what's going on here that it cannot find "powerline". Executing `:py import sys; print sys.path` seems to indicate that the right paths can be found, and I can execute `:py import powerline` without complaint.

For the record, I'm running the latest NeoVim (NVIM 0.0.0-alpha+201501262108) on Mac OS 10.10, with Python 2.7 installed through homebrew, and powerline 2.0 installed through pip.

ZyX-I:
@abstiles `powerline` that is not found has nothing to do with `powerline` module. It is an instance of `powerline.vim.VimPowerline` class that is expected to live in `__main__` module. If

    python foo = 1

will add variable `foo` to something *other* then `__main__` module bindings will not work. But in any case you should not try to use bindings for old Vim because the very reason `pyeval` was added in first place is that other ways around are fucking slow. NeoVim is using msgpack over sockets and this is going to be way slower. I was not going to add NeoVim support any time soon because this needs a major rewrite of Vim bindings since

```
&statusline
   \/
pyeval
   \/
…
   \/
segment function
   \/
vim.eval
   \/
result
   \/
displayed statusline
```


is fine, but

```
&statusline
  \/
:python          --------msgpack-rpc------>  … > segment function
                                                \/
vim_eval         <-------msgpack-rpc-------  vim.eval
  \/
eval result (C)  --------msgpack-rpc------>  result (Python)
                                                \/
displayed stl    <-------msgpack-rpc-------  segment function result
```

(in all cases: repeated for each segment function which bothers to use some interface with Vim) is not and I do not need tests to say that this would not work.

The only alternative approach I can see is

```
startup notification  -------msgpack-rpc------->  powerline
                                                     \/
                                                     …
                                                     \/
[...]

PR #2000, opened 2019-04-17
Defensive decode of vim.options['encoding'] in py3

Code0x58:
This change is defensive against the possibility that in some versions/environments the encoding is bytes instead of a string.

This is in response to the following exception (also seen in #1999):
```
  File "<string>", line 53, in <module>
  File "<string>", line 40, in powerline_troubleshoot
  File "/usr/lib/python3/dist-packages/powerline/lint/__init__.py", line 12, in <module>
    from powerline.segments.vim import vim_modes
  File "/usr/lib/python3/dist-packages/powerline/segments/vim/__init__.py", line 16, in <module>
    from powerline.bindings.vim import (vim_get_func, getbufvar, vim_getbufoption,
  File "/usr/lib/python3/dist-packages/powerline/bindings/vim/__init__.py", line 44, in <module>
    vim_encoding = get_vim_encoding()
  File "/usr/lib/python3/dist-packages/powerline/bindings/vim/__init__.py", line 25, in get_vim_encoding
    return vim.options['encoding'].decode('ascii') or 'ascii'
AttributeError: 'str' object has no attribute 'decode
```

Which might just be down to using neovim, as I haven't tested on vim with python3

### DIFF
--- a/powerline/bindings/vim/__init__.py
+++ b/powerline/bindings/vim/__init__.py
@@ -22,7 +22,8 @@ if (
 			return vim.options['encoding'] or 'ascii'
 	else:
 		def get_vim_encoding():
-			return vim.options['encoding'].decode('ascii') or 'ascii'
+			encoding = vim.options['encoding'] or 'ascii'
+			return encoding if isinstance(encoding, str) else encoding.decode('ascii')
 elif hasattr(vim, 'eval'):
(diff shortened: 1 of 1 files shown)


Code0x58:
Closing as there's much more to the problem which is really the use of neovim #1287

PR #2006, opened 2019-05-18
Extend documentation for bash

foursixnine:
When using a system that is updated frequently the user could find the situation where the path to the repository root has changed but the python binary is looking somewhere else (e.g, Python gets an upgrade from 3.6 to 3.7) or if the same bashrc is shared among different machines with different versions of python (2.x, 3.x and so on), is useful to be able to get one consistent way to figure out
what would be the path.

### DIFF
--- a/docs/source/usage/shell-prompts.rst
+++ b/docs/source/usage/shell-prompts.rst
@@ -39,6 +39,14 @@ absolute path to the Powerline installation directory (see :ref:`repository root
 
     in the bash configuration file. Without ``POWERLINE_BASH_*`` variables PS2 
     and PS3 prompts are computed exactly once at bash startup.
+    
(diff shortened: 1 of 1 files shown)


PH111P:
I really like this trick, however, as far as I understand it, it only works for local (per user) installations of the powerline. For global installations, `PYTHON_SITE_PATH=$(python -c "import site; print(site.getsitepackages()[0])")` should do the trick.

Slightly unrelated, for users of fish, the commands should look like 
```
set PYTHON_SITE_PATH (python -c "import site; print(site.getsitepackages()[0])")
source "$PYTHON_SITE_PATH/powerline/bindings/fish/powerline-setup.fish"
```

foursixnine:
@PH111P I just realized that the wording is a bit confusing... Will improve it soonish. 

Anything else needed?

Issue #1958, opened 2018-10-25
Git branch name mismatch between absolute path and symlinked path to same project directory


mtrp12:
I had a local git project. Say, in /home/user/projectdir. I moved this dir to /media/user/projectdir and created a symlink /home/user/projectdir -> /media/user/projectdir. Now it is like projectdir is still under home, although it is not. Now suppose, we were in master branch in the project. Now if I go to /home/user/projectdir and checkout another branch, say feature, powerline shows it correctly when I am on /home/user/projectdir. But it still shows master, if I go through /media/user/projectdir. I checked the .git/HEAD content, and both of the cases it was pointing to feature branch. Is it suppose to behave this way?

BlitzKraft:
That's a strange bug. I am able to reproduce this. Restarting powerline-daemon is a work-around. I need to look into it further but it might be caching issue.

PR #2031, opened 2019-10-15
Resolve symlink before getting branch name #1958

BlitzKraft:
Tests are missing. Need to write a try/catch for versions of python pre 3.4.

### DIFF
--- a/powerline/lib/vcs/git.py
+++ b/powerline/lib/vcs/git.py
@@ -87,7 +87,7 @@ class GitRepository(object):
 	def branch(self):
 		directory = git_directory(self.directory)
 		head = join(directory, 'HEAD')
-                # Resolve symlinks when determining branch to avoid cache conflicts issue#1958
+		# Resolve symlinks when determining branch to avoid cache conflicts issue#1958
 		if Path(directory).is_symlink():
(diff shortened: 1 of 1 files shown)


powerline/lib/vcs/git.py
```diff
@@ -86,6 +87,9 @@ def status(self, path=None):
 	def branch(self):
 		directory = git_directory(self.directory)
 		head = join(directory, 'HEAD')
+		# Resolve symlinks when determining branch to avoid cache conflicts issue#1958
+		if Path(directory).is_symlink():
```
PH111P: I think it would be better to put these kind of functions in https://github.com/powerline/powerline/blob/develop/powerline/lib/path.py

BlitzKraft: Makes sense. I'll modify that.

BlitzKraft:
I misspoke. The issue is not fixed yet.  I made a mistake when testing it on my system and (wrongly) concluded it's fixed.

BlitzKraft:
Seems I messed up rebasing. Need to squash merge this one, I think.

AronNovak:
This is still relevant.

PR #2032, opened 2019-10-17
Add nord colorscheme

ARifleman:
I was missing my favourite colorscheme that I use for everything, so I added it for my own use and figured I would share. I've been using it as my fish prompt for a bit and it seems to be working well. Please let me know if I'm missing anything important.

https://www.nordtheme.com/


PH111P:
I don't really have an opinion on this. 
On the one hand, having an additional default theme available is nice to have; on the other hand, I suppose that most users configure their powerline to use different colors anyways and adding another default theme incurs (slightly) more work when adding new highlight groups (for new segments).

PH111P:
Also, please fix `powerline-lint` errors.

Issue #1973, opened 2019-01-08
Yahoo weather APIs EOL


suoto:
Noticed that the weather segments disappeared, checked powerline log which contained

    2019-01-08 12:28:22,330:ERROR:tmux:WeatherSegment:Exception while computing state for 'London,United Kingdom,GB': <urlopen error [Errno -5] No address associated with hostname>

It seems that the URL powerline uses [is no longer supported by Yahoo since
03/Jan](https://developer.yahoo.com/weather/):

> Important EOL Notice: As of Thursday, Jan. 3, 2019, the weather.yahooapis.com and query.yahooapis.com for Yahoo Weather API will be retired. 
>
> To continue using our free Yahoo Weather APIs, use https://weather-ydn-yql.media.yahoo.com/forecastrss. Contact yahoo-weather-ydn-api@oath.com for credentials to onboard to this free Yahoo Weather API service.

For the record, here's the relevant section of the log 

    2019-01-08 12:28:22,330:ERROR:tmux:WeatherSegment:Exception while computing state for 'London,United Kingdom,GB': <urlopen error [Errno -5] No address associated with hostname>
    Traceback (most recent call last):
      File "/home/souto/.local/lib/python3.6/site-packages/powerline/lib/threaded.py", line 197, in render
        update_state = queries[key][1]
    KeyError: 'London,United Kingdom,GB'

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/lib/python3.6/urllib/request.py", line 1318, in do_open
        encode_chunked=req.has_header('Transfer-encoding'))
      File "/usr/lib/python3.6/http/client.py", line 1239, in request
        self._send_request(method, url, body, headers, encode_chunked)
      File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
        self.endheaders(body, encode_chunked=encode_chunked)
      File "/usr/lib/python3.6/http/client.py", line 1234, in endheaders
        self._send_output(message_body, encode_chunked=encode_chunked)
      File "/usr/lib/python3.6/http/client.py", line 1026, in _send_output
        self.send(msg)
      File "/usr/lib/python3.6/http/client.py", line 964, in send
        self.connect()
      File "/usr/lib/python3.6/http/client.py", line 936, in connect
        (self.host,self.port), self.timeout, self.source_address)
      File "/usr/lib/python3.6/socket.py", line 704, in create_connection
        for res in getaddrinfo(host, port, 0, SOCK_STREAM):
[...]

PR #2038, opened 2019-10-26
Shifted away from (abandoned) Yahoo API to OpenWeatherMap

StopMotionCuber:
I made some efforts to make the Weather segment work again.
This adresses issue #1973
Downside of OpenWeatherMap is that it needs an API key, but they state [to welcome FOSS](https://openweathermap.org/faq) so I'm going to request an API key for this project there.

This is not really ready for merging yet, but already in a working state (if you provide an API key).
I'm not gonna lie, I submitted this PR early as hacktober is almost over and this is my last PR that I need for Hacktoberfest.


ingorichter:
I'm interested in getting this for myself. What is the best way to get this PR merged?

StopMotionCuber:
Bumping this issue was a good idea to get this MR rolling.
I think this can be merged now, although I did not take a too deep look into the icons associated with weather codes returned by OpenWeatherMap. Right now the icons selected are rather generic.

ingorichter:
How do you get the icons? I see only the textual representation of the current weather. But it's fine for me. Did you consider mapping all different weather situations from https://openweathermap.org/weather-conditions? That would be a huge improvement in detail information provided. The PR looks good 👏

StopMotionCuber:
The icons are applied by the skin based on the textual representation.
You can basically take a look at the diff for the `weather_conditions_codes` dict which names are available.
When there are multiple names, the skin will try to match the first name, in case there is no icon it will fall back to the generic name (at least that is my understanding).

powerline/segments/common/wthr.py
```diff
@@ -105,6 +105,7 @@ class WeatherSegment(KwThreadedSegment):
 	interval = 600
 	default_location = None
 	location_urls = {}
+	weather_api_key = "fbc9549d91a5e4b26c15be0dbdac3460"
```
PH111P, line 108: Perhaps I'm missing something here, but it seems wrong to me to have an API key hard-coded.

ingorichter: that could well be in the config file :+1

ingorichter:
@StopMotionCuber I was thinking Icon==Image

powerline/segments/common/wthr.py
StopMotionCuber: Yeah, I will  to make that work.

StopMotionCuber:
> @StopMotionCuber I was thinking Icon==Image

I'm not sure what exactly you mean. As we are on the command line, images are generally not possible, at least not in a png way. This segment works with Unicode symbols/emojis.
If you mean something else please clarify that, as I'm not sure whether I'm getting your point.

StopMotionCuber:
You can find one example for the unicode symbol config here:
https://github.com/powerline/powerline/blob/develop/powerline/config_files/themes/powerline_unicode7.json#L73

ingorichter:
@StopMotionCuber :-) Yes, I was thinking symbols/emojis is the only way to show something on the command line, that's why I was wondering about the term icon in this context. Thanks for the explanations. I just wonder why I don't see the emojis in my setup 🤔 Instead of the clouds emoji I see **CLOUDS** at string instead

powerline/segments/common/wthr.py
PH111P: How about adding the api key as a parameter to the segment?

StopMotionCuber:
I changed the structure a bit, now the API key can be changed and it's also documented how to change.
I took a look at the IMAP segment for this implementation

PH111P:
Looks good! Thanks for the great work @StopMotionCuber!
I suppose this PR is now ready for merging (once the tests pass, which they will, I presume)

StopMotionCuber:
Quick update on this, I just got a reply from the OpenWeatherMap support and the API key is now registered under opensource conditions (3000 API calls/min)

PR #2048, opened 2019-11-19
Update Powerline for ipython >=7

vegerot:
The method used for `since_5` does not work, but `since_7` does, however it is not documented.

I personally haven't found `c = get_config()` to be useful once you have generated the default `~/.ipython/profile_default/ipython_config.py` file.

Since `c.TerminalInteractiveShell.simple_prompt` is set to `False` by default, it's not worth mentioning it.  There's many other things that **can** break if you change the default config options too.  Additionally, `simple_prompt=True` doesn't fatally break anything— just make things less pretty.  So it'd be better to not have that, but mention if things don't look pretty to make sure they didn't change that default.

Additionally updated `post_0_11.py` to use `since_7` (given that most people using a later version than 0.11 are probably using 7), and instructing users with issues (for example, those using version 5) to read the documentation for `since_5`.

### DIFF
--- a/docs/source/usage/other.rst
+++ b/docs/source/usage/other.rst
@@ -130,6 +130,46 @@ root <repository-root>`)::
 IPython prompt
 ==============
 
+For IPython>=7.0, add the following line to
+:file:`~/.ipython/profile_default/ipython_config.py` file in the used profile:
+
+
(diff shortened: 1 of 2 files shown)


PH111P:
Nice, thanks!

PR #2057, opened 2019-12-01
Add vim plugin coc.nvim segments.

### LINKED ISSUE #2056: Add vim plugin coc.nvim status segment
May we could think of supporting display vim plugin coc.nvim compile status (Error number and Warning number), like vim ale segment do.

chrisniael:
Close #2056

### DIFF
--- a/powerline/segments/vim/plugin/coc.py
+++ b/powerline/segments/vim/plugin/coc.py
@@ -1,0 +1,47 @@
+# vim:fileencoding=utf-8:noet
+from __future__ import (unicode_literals, division, absolute_import, print_function)
+
+try:
+	import vim
+except ImportError:
+	vim = object()
+
+from powerline.bindings.vim import vim_command_exists
+from powerline.theme import requires_segment_info
+
+# coc_status's format: E1 W2
+def parse_coc_status(coc_status):
+	# type(coc_status) is tuple
+	errors_count = 0
+	warnings_count = 0
+	if len(coc_status) <= 0:
+		return errors_count, warnings_count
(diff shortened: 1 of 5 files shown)


PH111P:
Hi @chrisniael,

Nice work. Just a small thing: The tests seem to fail as you didn't add default colors for the new highlight groups defined in your segment:
```
Error while checking theme (key /segments/left/list item 12/function)
found highlight group coc not defined in the following colorschemes: default, solarized, solarizedlight
```

I'd be nice if you could fix this.

powerline/segments/vim/plugin/coc.py
```diff
@@ -0,0 +1,47 @@
...
+	for item in status_list:
+		if len(item) > 0 and item[0] == 'E':
+			errors_count = int(item[1:])
+		if len(item) > 0 and item[0] == 'W':
+			warnings_count = int(item[1:])
+	return errors_count, warnings_count
+
+@requires_segment_info
+def coc(segment_info, pl):
+	segments = []
```
PH111P: The documentation for this function is missing: `powerline-lint` does not know how the highlight groups are named. Try adding:
```
    '''<add description here>

    Highlight groups used: ``coc:warning`` or ``warning``, ``coc:error`` or ``error``.
    '''
```

chrisniael: Thanks for the tip.

giovanniiapaolo:
I'd like to implement this in my setup, but I just can't seem to get it. I'm not even sure if it's available in the latest apt

PR #2062, opened 2019-12-11
Use updated i3ipc syntax in i3 segments/listers

### LINKED ISSUE #2046: Breaking change in python-i3ipc breaks i3 listers
A relatively recent, major update in i3ipc (`v2.0.1`) breaks some of our segments and listers here.
I discovered it today while updating all my python packages (for python 3.8).

In essence, the python class wrappers around i3ipc JSON responses that our code deals with have changed, so that `reply['attr']` is no longer valid, and `reply.attr` or `getattr(reply, 'attr')` must be used.
This is relevant in powerline/listers/i3wm.py and powerline/segments/i3wm.py.

s-ol:
Close #2046

### DIFF
--- a/powerline/segments/i3wm.py
+++ b/powerline/segments/i3wm.py
@@ -12,11 +12,11 @@ WORKSPACE_REGEX = re.compile(r'^[0-9]+: ?')
 
 def workspace_groups(w):
 	group = []
-	if w['focused']:
+	if w.focused:
 		group.append('w_focused')
-	if w['urgent']:
+	if w.urgent:
 		group.append('w_urgent')
-	if w['visible']:
+	if w.visible:
 		group.append('w_visible')
 	group.append('workspace')
 	return group
@@ -52,12 +52,12 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
 
 	return [
 		{
-			'contents': w['name'][strip:],
+			'contents': w.name[strip:],
 			'highlight_groups': workspace_groups(w)
 		}
 		for w in get_i3_connection().get_workspaces()
-		if ((not only_show or any(w[typ] for typ in only_show))
-		    and (not output or w['output'] == output))
+		if ((not only_show or any(getattr(w, typ) for typ in only_show))
+		    and (not output or w.output == output))
 	]
 
 
@@ -80,7 +80,7 @@ def workspace(pl, segment_info, workspace=None, strip=False):
 		try:
 			w = next((
 				w for w in get_i3_connection().get_workspaces()
-				if w['name'] == workspace
+				if w.name == workspace
 			))
 		except StopIteration:
 			return None
@@ -90,13 +90,13 @@ def workspace(pl, segment_info, workspace=None, strip=False):
 		try:
 			w = next((
 				w for w in get_i3_connection().get_workspaces()
-				if w['focused']
+				if w.focused
 			))
 		except StopIteration:
 			return None
 
 	return [{
-		'contents': format_name(w['name'], strip=strip),
+		'contents': format_name(w.name, strip=strip),
 		'highlight_groups': workspace_groups(w)
 	}]
 
@@ -150,6 +150,6 @@ def scratchpad(pl, icons=SCRATCHPAD_ICONS):
 			'contents': icons.get(w.scratchpad_state, icons['changed']),
 			'highlight_groups': scratchpad_groups(w)
 		}
-		for w in get_i3_connection().get_tree().descendents()
+		for w in get_i3_connection().get_tree().descendants()
 		if w.scratchpad_state != 'none'
 	]
--- a/docs/source/develop/segments.rst
+++ b/docs/source/develop/segments.rst
@@ -528,10 +528,10 @@ i3wm
     in lemonbar bindings.
 
 ``workspace``
-    dictionary containing the workspace name under the key ``"name"`` and 
-    boolean values for the ``"visible"``, ``"urgent"`` and ``"focused"`` 
-    keys, indicating the state of the workspace. Currently only provided by 
-    the :py:func:`powerline.listers.i3wm.workspace_lister` lister.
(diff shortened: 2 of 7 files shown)


powerline/segments/i3wm.py
```diff
@@ -80,23 +83,26 @@ def workspace(pl, segment_info, workspace=None, strip=False):
...
+				if w.name == workspace
 			))
 		except StopIteration:
 			return None
 	elif segment_info.get('workspace'):
-		w = segment_info['workspace']
+		w = Dummy()
+		info = segment_info['workspace']
+		for k in info:
+			setattr(w, k, info[k])
```
s-ol: kind of an eyesore but what can you do... I tried `setattr` on the dict itself, but since it's a CPython object that doesn't work.

PH111P: Personally, I use the following to obtain a new dummy workspace, perhaps something similar could work here?
```
import copy
def get_next_ws(ws, outputs):
    names = [w.name for w in ws] 
    for i in range(1, 100):
        if not str(i) in names:
            res_ls = []
            res = copy.deepcopy(ws[0])
            res.num = -5
            res.name = str(i)
            res.urgent = False
            res.focused = False
            res.visible = False
            for o in outputs:
                r2 = copy.deepcopy(res)
                r2.output = o 
                res_ls += [r2]
            return res_ls
    return []
```

PH111P:
Currently, the tests seem to fail; relevant output (could be related to your `Dummy` hack):
```
======================================================================
ERROR: test_workspace_lister (__main__.TestI3WM)
----------------------------------------------------------------------
  File "/home/travis/build/powerline/powerline/tests/test_python/test_listers.py", line 44, in test_workspace_lister
    list(i3wm.workspace_lister(pl=pl, segment_info={'a': 1})),
  File "/home/travis/build/powerline/powerline/powerline/listers/i3wm.py", line 67, in <genexpr>
    and (not output or w.output == output)))
AttributeError: 'dict' object has no attribute 'output'

======================================================================
ERROR: test_scratchpad (__main__.TestI3WM)
----------------------------------------------------------------------
  File "/home/travis/build/powerline/powerline/tests/test_python/test_segments.py", line 1076, in test_scratchpad
    self.assertEqual(i3wm.scratchpad(pl=pl), [
  File "/home/travis/build/powerline/powerline/powerline/segments/i3wm.py", line 159, in scratchpad
    for w in get_i3_connection().get_tree().descendants()
AttributeError: 'Conn' object has no attribute 'descendants'
======================================================================
ERROR: test_workspace (__main__.TestI3WM)
----------------------------------------------------------------------
  File "/home/travis/build/powerline/powerline/tests/test_python/test_segments.py", line 1029, in test_workspace
    self.assertEqual(i3wm.workspace(pl=pl, segment_info=segment_info, workspace='1: w1'), [
  File "/home/travis/build/powerline/powerline/powerline/segments/i3wm.py", line 85, in workspace
    w for w in get_i3_connection().get_workspaces()
  File "/home/travis/build/powerline/powerline/powerline/segments/i3wm.py", line 86, in <genexpr>
    if w.name == workspace
AttributeError: 'dict' object has no attribute 'name'
======================================================================
ERROR: test_workspaces (__main__.TestI3WM)
----------------------------------------------------------------------
  File "/home/travis/build/powerline/powerline/tests/test_python/test_segments.py", line 981, in test_workspaces
    self.assertEqual(i3wm.workspaces(pl=pl, segment_info=segment_info), [
  File "/home/travis/build/powerline/powerline/powerline/segments/i3wm.py", line 60, in workspaces
    and (not output or w.output == output))
AttributeError: 'dict' object has no attribute 'name'
```

powerline/segments/i3wm.py
s-ol: Hm, wondering whether there is a use case where we would like that segment to work *without* a connection to i3, because it seems like the Dummy class currently allows that where your solution wouldn't.

PH111P: You could probably update my solution to also add dummy values for the output (and perhaps the other things that a workspace object should have); then you wouldn't need a connection anymore (but the code would become even more ugly).

s-ol:
The upper mistakes I think are legit issues I introduced, in code paths that I just didn't exercise in my quick testing. In general `dict`s should be floating around after `workspace_groups` is called, while before the other wrappers are used.

s-ol:
I would say that is because the test wrapper creates `dict`s as test fixtures, which is not what i3ipc generates now.

s-ol:
Thought about this again... we need a better solution if we want to keep supporting `i3-py` as an alternative binding; `i3-py` returns dicts. So maybe we need some kind of compatibility layer like a `get_attr_or_index` that does the corresponding action, and rewrite the code in terms of that?

PH111P:
Perhaps we should just drop support for `i3-py`? It has been deprecated for a while now, and there is a drop-in replacement available.

s-ol:
also a fair choice

s-ol:
With the python 3.9 release this once again came to my attention and I finally sat down and did it right (including tests). As we decided above, no more i3-py. The `segment_info` now contains the "raw" workspace object from the connection, so there are no more dicts involved (until the powerline segment output, of course).

powerline/segments/i3wm.py
```diff
@@ -52,15 +52,18 @@ def workspaces(pl, segment_info, only_show=None, output=None, strip=0):
...
 		}
 		for w in get_i3_connection().get_workspaces()
-		if ((not only_show or any(w[typ] for typ in only_show))
-		    and (not output or w['output'] == output))
+		if ((not only_show or any(getattr(w, typ) for typ in only_show))
+		    and (not output or w.output == output))
 	]
 
 
+class Dummy:
```
PH111P, line 64: Why do we need this `Dummy` class?

s-ol: ah, I knew I should've started from scratch ;)

PH111P (approved):

PR #2095, opened 2020-05-23
Update to include default path to powerline_root for apt-get

ymode:
This isn't super important and would be considered a low priority change - however there is a lot of people googling for the same answer that this one line could cover.


PH111P:
Thanks!

PR #2096, opened 2020-05-25
Added timezone support to time

rdelfin:
This adds support in the time segments to take in optional timezone and suffix parameters. This will allow users to display multiple time zones  on their powerline. This is especially useful on servers that are configured to have a standard timezone for ease of configuration (like UTC) but the user will be using the machine from a specific local timezone.

### DIFF
--- a/powerline/segments/common/time.py
+++ b/powerline/segments/common/time.py
@@ -6,7 +6,7 @@ from datetime import datetime
 import pytz
 
 
-def date(pl, format='%Y-%m-%d', istime=False, timezone=None, suffix=None):
+def date(pl, format='%Y-%m-%d', istime=False, timezone=None):
 	'''Return the current date.
 
 	:param str format:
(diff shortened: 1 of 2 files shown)


rdelfin:
I'm not sure this is ready for merge. Part of my issue is that, when testing locally, adding the timezone arg in my theme config makes the segment go away.  Interestingly enough though, running `powerline-render tmux right` does show the additional segment. Does anyone know why there would be this discrepancy? I can confirm that removing the "timezone" arg makes the segment immediately re-appear on my powerline.

PH111P:
Hi, 
While I appreciate your work, I think there is a simpler solution using strptime (that doesn't require pytz):

``` python
    tz = datetime.strptime(timezone, '%z').tzinfo if timezone else None
```

I have such an implementation lying around; perhaps now is a good time to polish it.

PH111P:
Also see https://github.com/powerline/powerline/pull/1337

PH111P:
> I'm not sure this is ready for merge. Part of my issue is that, when testing locally, adding the timezone arg in my theme config makes the segment go away. Interestingly enough though, running `powerline-render tmux right` does show the additional segment. Does anyone know why there would be this discrepancy? I can confirm that removing the "timezone" arg makes the segment immediately re-appear on my powerline.

Regarding this discrepancy, try to restart the powerline-daemon. The old daemon still doesn't know about the new argument to the segment and thus errors (and thus cannot compute the segment).

PR #2112, opened 2020-07-24
Update linux.rst

kesavanvt:
conf.d is not a directory


PH111P:
While the directory `~/.config/fontconfig/conf.d/` may not exist initially, it should indeed be a directory and not a file; font configuration files are then supposed to go into that directory (as `~/.config/fontconfig/conf.d/NN-name.conf`). So the documentation is indeed correct as-is.

Issue #2101, opened 2020-05-31
Tmux + fish returns wrong cwd


Pingger:
Hi,

I thought I'd migrate from tmux-powerline, but it seems there is a regression from tmux-powerline:
When using `fish` inside a `tmux` session, the tmux-powerline fails to detect changes in the `cwd`. This causes all segments requiring the `cwd` to break. (like `powerline.segments.common.env.cwd`, `powerline.segments.common.vcs.branch`,...)

Any idea what I can do to fix this issue?
- No not using `tmux` is not an option
- No not using `fish` is not an option

I'm using powerline from the debian package repositories. The installed version is **2.7.3**
fish and tmux are also from the debian package repositories with versions **3.1.2-1** and **3.1b-1** (respectively)

Greetings.

PH111P:
Perhaps the discussion in https://github.com/powerline/powerline/issues/467 can yield some insights. As I understand it, the cwd segment currently just doesn't work in tmux directly.

Pingger:
So we have a workaround for >5 years ... 
The easy way would be to just use this: `#{pane_current_path}` this expands to the path of the current tmux pane. Unfortunately I don't really do python ... Throw java/c/bash or even haskell at me, but for some reason python and me don't really mix.

The command to be run within the tmux session would be `tmux display -p "#{pane_current_path}"`. which is the equivalent of `echo $PWD`

I hope that helps.

Greetings

abujalski:
@Pingger you can try this https://github.com/abujalski/powerline-segments

It works for me, hope it will work for you.

StopMotionCuber:
This can be closed as #2127 got merged

PR #2127, opened 2020-10-05
Fixed getcwd for tmux

### LINKED ISSUE #467: cwd segment not working in tmux
I dunno why, I have added the `cwd` segment, and the only thing I'm getting is the `~` sign.

I run `powerline tmux right` and the output it's okay (showing the cwd), but the status line in tmux doesn't change at all.

How can I continue debugging the issue?

StopMotionCuber:
It's hacktoberfest again, so I decided to go through the backlog of this project.
This fixes #467 (which for some reason is already closed, anyway it's from ancient times) as well as #2101

On another note, I wanted to ask whether we can drop support for ancient versions of tmux (prior to 2.1) and clean the bindings up here a bit.
The most recent version we are considering to support is from October 2015 (this one would still be supported, only older releases are not), and I don't think anybody will be mad if we drop support for tmux older than that.
If someone still uses a tmux version older than that, I doubt that they will update their powerline anyway.


StopMotionCuber:
Some reference how this works:
When we call the `powerline-render` process, `-R pane_current_path=#{pane_current_path}` got added now. This will be parsed and stored in the segment info.
If we find that item in the segment info, we use a `getcwd` function that returns exactly this path.

Issue #2128, opened 2020-10-05
Make unittests work outside of CI environment


StopMotionCuber:
Currently all the unit tests only work with CI environment.
As this is rather tedious to setup, I'd propose to change this test environment.
My goal there would be 
1. to make it work within PyCharm
2. to not change any behaviour if we are in CI environment.

Background: When working on update for weather segment, I noticed how awful it is, to always wait 10 minutes and scroll through 5k lines, just because there is some little error in the your segements or in the tests for them.
There I already temporarily achieved to make it work with PyCharm (as I didn't commit those changes, I didn't care about CI, but respecting that should also be possible).

I'd happily address this issue myself, as I have already taken a look at this.

Also, do we have some further things that should be addressed within the test environment?
If wanted, I may take a look at running the tests with `pytest` so that the 5k-lines-nobody-finds-where-the-actual-issue-is-output may become somewhat readable

StopMotionCuber:
Aaaargh, I'm trying to transition to pytest + GitHub actions. While _most_ tests are working, we have a few racy tests which need a responsive machine which is why they tend to fail on actions (I assume cause we don't get enough CPU share, either by policy or cause they are running on machines with a high load).
[Here](https://github.com/StopMotionCuber/powerline/actions) is a link if someone cares, I'll open a PR once I get tests to somewhat pass

PR #2131, opened 2020-10-13
WIP: Test improvements

StopMotionCuber:
This is a WIP Pull request which directly addresses #2128
Let me elaborate the idea that I have for the testing environment and the changes that I have done:

- Move from Travis CI to Github Actions (plan is to ditch Travis CI completely once the PR is feature-complete
  - Some tests were rewritten, as Github Actions seems to have slow executors, which lead to failure of racy tests
- Execute python tests with [pytest](https://docs.pytest.org/en/stable/)
  - Uses a few threads -> tests testing on threads were changed
  - COLORS!!!
  - Verbose output on failure, compact output on success
- Rewrite ALL tests that are currently done using some weird custom shell test environment to also use pytest.
  - Awesome tests have already been rewritten
  - Processes that are tested, are executed as separate processes
  - This also provides readable output
- Make a single sections in GitHub Actions for each test suite, this provides a better overview which suite exactly failed

Please let me know if you have some general concerns about the style of this test suite. If this is not the case, I'll continue rewriting the missing tests.

PS: I'd appreciate getting a `hacktoberfest-accepted` label for this PR once it is feature-complete


PH111P:
This looks great! (Especially nice is the fact that the tests work for python 3.7 to 3.9)

Somehow, the test produce a different results for almost every python version (which I find somewhat strange), but I guess this is due to the WIP nature of the PR. In any case, you may drop test support for python 2.7 and  ≤ 3.5; would it be possible to test pypy versions?

StopMotionCuber:
I basically dropped support for Python 2, so you can ignore that, that will be removed.
3.5 is a pathlib issue, I'll fix that one.
Seems like the environment is not 100% reproducable which leads to different renders. I'll try to fix that (but works on my machine with 3.8).
Probably GitHub is using different base images for 3.6 and 3.7
Quick googling showed that pypy should also be supported with GitHub actions, so I'll also add that one.

StopMotionCuber:
Some questions about the shell tests:
I did not manage to get busybox working fully, as there seemes to be some weird stuff going around for the jobnumber, which will show the actual job number too late, therefore the test is not passing. I could reproduce this by hand.
The behaviour is that when executing `bgscript.sh & waitpid.sh` the jobnumber 1 is only shown after pressing enter twice, not once how you would expect. How to proceed here?

Fish is also not passing, but this was not included for Travis CI tests before, so maybe something is diverging there. Can I just drop that one from the tests?

PH111P:
I guess for now it would be OK to drop these tests; perhaps we should look into them again later, though.

destinehacker:
its a fake code sorry 8cee18dbe56c0b0232749f57ced628503cd2e0d3

Antiz96:
Hi,

We are trying to [stop using `python setup.py test`](https://archlinux.org/todo/stop-using-python-setuppy-test/) in packages on Arch Linux side. I understand that the changes addressed by this PR are representing a huge effort but, since it didn't get much attention since a few years now, I was wondering if the effort is still ongoing today and if we can expect a switch to `pytest` at some point?

Thanks a lot for your work!

StopMotionCuber:
Oh well, it's been a long time since I worked on it. It's not really high on my radar right now and I don't really use Powerline itself anymore, but am rather going for alternatives. Could take a look how the state is right now if I find time for that

PR #2134, opened 2020-10-26
Fix non-meaningful venv names.

### LINKED ISSUE #2094: Segment virtualenv isn't always as useful as I'd like...
I work on a lot of python projects, and a general standard with all of them seems to be a $PROJECT/.venv directory for that's $PROJECT's virtualenv. I move between projects a lot, and have to remember to `deactivate; activate` each time I do. Which of course I sometimes forget to do...

Unfortunately this means that my prompt is always:

`ash > (e) .venv > ~ > src > project`

Which has left me sometimes wondering which of the .venv's is active (since I've got like 6 of them.) And so I'm always second guessing myself about if I've changed or not.

Ideally, I think it'd be nice if instead of just the raw basename of the env directory being the segment, it was nice enough to show one directory level higher up if the basename is `.venv`. This way I'd get `(e) project` instead.

(Yes, I can, and likely will, write my own segment for this, but it seems a common enough use case to perhaps change the stock segment.)

StopMotionCuber:
Went a bit through the project backlog.
This fixes #2094 and includes tests for the additional behaviour introduced, namely going up a directory if a venv name is on a blacklist (which is useful if all your venv's are called `venv` and the meaningful name is the project dir above).
This list can be configured, currently the blacklist is `venv` and `.venv`.

I doubt that there are concerns with these 2 names, however if someone wants to have additional default venv names, I can also include them.

PS: I'd appreciate getting a `hacktoberfest-accepted` label for this PR

### DIFF
--- a/powerline/segments/common/env.py
+++ b/powerline/segments/common/env.py
@@ -21,7 +21,7 @@ def environment(pl, segment_info, variable=None):
 @requires_segment_info
 def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False, ignored_names=("venv", ".venv")):
 	'''Return the name of the current Python or conda virtualenv.
-	:param Iterable ignored_names:
+	:param list ignored_names:
 		Names of venvs to ignore. Will then get the name of the venv by ascending to the parent directory
 	:param bool ignore_venv:
 		Whether to ignore virtual environments. Default is False.
@@ -36,6 +36,7 @@ def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False, ignored_
 		for candidate in reversed(segment_info['environ'].get('CONDA_DEFAULT_ENV', '').split("/")):
 			if candidate and candidate not in ignored_names:
 				return candidate
+	return None
 
 
 @requires_segment_info
--- a/powerline/segments/common/env.py
+++ b/powerline/segments/common/env.py
@@ -19,20 +19,23 @@ def environment(pl, segment_info, variable=None):
 
 
 @requires_segment_info
-def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False):
+def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False, ignored_names=("venv", ".venv")):
 	'''Return the name of the current Python or conda virtualenv.
-
+	:param Iterable ignored_names:
+		Names of venvs to ignore. Will then get the name of the venv by ascending to the parent directory
 	:param bool ignore_venv:
 		Whether to ignore virtual environments. Default is False.
 	:param bool ignore_conda:
 		Whether to ignore conda environments. Default is False.
 	'''
-	return (
-		(not ignore_venv and
-		 os.path.basename(segment_info['environ'].get('VIRTUAL_ENV', ''))) or
-		(not ignore_conda and
-		 segment_info['environ'].get('CONDA_DEFAULT_ENV', '')) or
-		None)
+	if not ignore_venv:
+		for candidate in reversed(segment_info['environ'].get('VIRTUAL_ENV', '').split("/")):
+			if candidate and candidate not in ignored_names:
+				return candidate
+	if not ignore_conda:
+		for candidate in reversed(segment_info['environ'].get('CONDA_DEFAULT_ENV', '').split("/")):
+			if candidate and candidate not in ignored_names:
+				return candidate
 
 
 @requires_segment_info
--- a/powerline/bindings/zsh/__init__.py
+++ b/powerline/bindings/zsh/__init__.py
@@ -101,10 +101,7 @@ class Environment(object):
 			return False
 
 
-if hasattr(getattr(zsh, 'environ', None), '__contains__'):
-	environ = zsh.environ
-else:
-	environ = Environment()
+environ = Environment()
 
 
 if hasattr(zsh, 'expand') and zsh.expand('${:-}') == '':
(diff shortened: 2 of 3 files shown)


powerline/segments/common/env.py
```diff
@@ -19,20 +19,23 @@ def environment(pl, segment_info, variable=None):
 
 
 @requires_segment_info
-def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False):
+def virtualenv(pl, segment_info, ignore_venv=False, ignore_conda=False, ignored_names=("venv", ".venv")):
```
PH111P, line 22: As a very minor comment, can you use a list instead of a tuple for the default argument? (This makes it similar to the rest of the code and is also less confusing when configuring the segment via json, where the param would become a list anyway.)

Other than that, nice work!

powerline/segments/common/env.py
```diff
@@ -19,20 +19,23 @@ def environment(pl, segment_info, variable=None):
...
-		 segment_info['environ'].get('CONDA_DEFAULT_ENV', '')) or
-		None)
+	if not ignore_venv:
+		for candidate in reversed(segment_info['environ'].get('VIRTUAL_ENV', '').split("/")):
+			if candidate and candidate not in ignored_names:
+				return candidate
+	if not ignore_conda:
+		for candidate in reversed(segment_info['environ'].get('CONDA_DEFAULT_ENV', '').split("/")):
+			if candidate and candidate not in ignored_names:
+				return candidate
```
PH111P, line 39: Also, I think it would be more in line with the rest of the code to explicitly return `None` here.

StopMotionCuber: I thought that was kind of redundant, but I can add that

powerline/segments/common/env.py
StopMotionCuber: PyCharm complains about that :disappointed:
Background: When the default argument is a list, it is mutable. This means that _if_ I would change that list within the function (e.g. append a value), this change is permanent. In this case when appending a value within the function this list would grow and grow.

Given that the code is rather small and the values are only read, this is not a problem _here_, but when requiring immutable, I think having an immutable object by default makes this whole thing more explicit

PH111P: Oh indeed, this is a very good point. Let's keep the tuple then.

powerline/segments/common/env.py
PH111P: While that is true, I think adding an explicit return statement makes the code more readable, as it pronounces the end of that function. (And even if it is redundant, it is only a single line, so it should be fine.)

powerline/segments/common/env.py
PH111P: In fact, I just looked again at the other parameters where a list is expected and they all seem to either lack a default argument or have it set to `None`, so having a tuple here is actually fine, I'd say. In any case, the documentation should say `param list ...` though.

PH111P:
Is there any reason why the tests fail? (In particular, the zsh tests seem to fail)

StopMotionCuber:
My guess would be "yes".
However, I cannot correlate that failure to the actual PR and I will definitely not debug it until the new test environment is ready (I think I already ranted enough about the testing environment)

PH111P:
Yes indeed, I expected such a reply (sorry for the bad question.)

Some more details from the test output:

```
--- /home/travis/build/powerline/powerline/tests/test_shells/outputs/zsh.zpython.ok	2020-10-26 23:38:13.122084117 +0000
+++ /home/travis/build/powerline/powerline/tests/tmp/shell//zsh.zpython.zpython.log	2020-10-26 23:42:41.227169087 +0000
@@ -2,7 +2,7 @@
   HOSTNAME  USER   BRANCH  …  tmp  shell  3rd  cd .git
   HOSTNAME  USER   BRANCH  …  shell  3rd  .git  cd ..
   HOSTNAME  USER   BRANCH  …  tmp  shell  3rd  VIRTUAL_ENV="/home/USER/.virtenvs/some-virtual-environment"
-  HOSTNAME  USER  (e) some-virtual-environment   BRANCH  …  tmp  shell  3rd  VIRTUAL_ENV=
+  HOSTNAME  USER   BRANCH  …  tmp  shell  3rd  VIRTUAL_ENV=
   HOSTNAME  USER   BRANCH  …  tmp  shell  3rd  bgscript.sh & waitpid.sh
 [1] PID
   HOSTNAME  USER   BRANCH  …  tmp  shell  3rd  1  false
```

Some more output, note that the error comes from the python 3.6 run:

```
  travis-job-ed4bc75f-88f7-4292-9696-9e366c36dd8c  travis   2e03695  …  build  powerline  powerline  2020-10-26 23:42:39,433:ERROR:shell:virtualenv:Exception while computing segment: a bytes-like object is required, not 'str'
  File "/home/travis/build/powerline/powerline/powerline/segment.py", line 173, in process_segment
    contents = segment['contents_func'](pl, segment_info)
  File "/home/travis/build/powerline/powerline/powerline/segment.py", line 412, in <lambda>
    contents_func = lambda pl, segment_info: _contents_func(pl=pl, segment_info=segment_info, **args)
  File "/home/travis/build/powerline/powerline/powerline/segments/common/env.py", line 32, in virtualenv
    for candidate in reversed(segment_info['environ'].get('VIRTUAL_ENV', '').split("/")):
TypeError: a bytes-like object is required, not 'str'
```

I think that fixing this is worth it even before the new tests are ready.

StopMotionCuber:
Thanks for isolating the traceback, this makes the issue debuggable for me.

So this seems to be caused by https://github.com/powerline/powerline/blob/develop/powerline/bindings/zsh/__init__.py#L105 where we might get byte-like objects in our environment (Then we would need to call `split(b'/')`).
Also not _all_ zsh tests are failing, but only the `zpython` ones.

As I couldn't find the sources for `zpython` online (you can find an issue referencing a dead bitbucket of `zpython`/`libzpython` here https://github.com/Homebrew/legacy-homebrew/issues/40318) I guess that project is dead. @ZyX-I is the maintainer of `zpython`, so maybe he can comment on that one.

I don't think it's worth fixing a dependency that is _that_ dead that not even the original repository is available

StopMotionCuber:
Also, if you ask, Travis will probably get it's zpython from https://github.com/powerline/deps/tree/master/zpython

PH111P:
Thanks for looking into this! I think it is OK to just drop zpython then.

PR #2148, opened 2021-01-02
Emulate a right prompt in bash

### LINKED ISSUE #2103: How hard would it be to have a right prompt in bash?
According to [this](https://superuser.com/questions/187455/right-align-part-of-prompt/1203400#1203400), one could implement a right hand prompt for bash. 

How hard would it be to have powerline produce sth like that? 

What part of the code should I need to look into if I tried to add support for sth like that?

sdrik:
Fixes #2103


PH111P:
This looks very nice! 
(Although, I must admit that it does not look as "trivial" as I would have imagined it to be, so it appears I was wrong with my assessment in that regard; so all the more, thanks for your contribution!)

Some comments:
* Allowing to pass arguments from `render` to `hl` and `hlstyle` is probably a good idea, so although this makes this PR look slightly messy, it is a good change in my opinion. However, please add the argument (`hl_args`) to the corresponding documentation as well. (In particular in `renderer.py`.)

Issue #2158, opened 2021-03-10
Different coloured Zoom indicator


cibinmathew:
**Feature Request**:
Show a more noticeable zoom indicator (https://github.com/powerline/powerline/pull/1436) may be in different color or font icon.
![image](https://user-images.githubusercontent.com/10793628/110615231-34d15900-81b9-11eb-8a15-59da294dd980.png)

PR #1436, opened 2015-08-26
Adding zoom indicator to window status formats.

setrofim:

PR #2180, opened 2021-09-28
Added zoom indicator icon

cibinmathew:
Shows a more noticeable zoom indicator (#2158 #1436)

image


PH111P:
Hi @cibinmathew , 
Thanks for your work, a zoom indicator seems like a very good idea. However, some things:
* Hard-coding a non-ASCII character is not very portable (and in particular may output weird things in a terminal). Perhaps it would be better to find a way to configure the appearance.
* Although the CI is currently broken (:cry:), the tests would need to be updated as well.

cibinmathew:
> Hi @cibinmathew , Thanks for your work, a zoom indicator seems like a very good idea. However, some things:
> 
> * Hard-coding a non-ASCII character is not very portable (and in particular may output weird things in a terminal). Perhaps it would be better to find a way to configure the appearance.
> * Although the CI is currently broken (cry), the tests would need to be updated as well.

@PH111P 
1. Is the preferred location to store configs `powerline/config.json` or I can try non ascii icon configurable using env variable
2. I am unable to see the test results. Is there a command to trigger the tests in local to debug.

cibinmathew:
> Please see https://powerline.readthedocs.io/en/latest/configuration.html for the configuration structure. If I'm not mistaken, the flow should be roughly as follows: The config sets the overall theme (non-ascii / unicode / etc); then depending on the config, the tmux renderer sets some environment variable (such as currently `$_POWERLINE_WINDOW_CURRENT_COLOR` and so on), which are then parsed by tmux for the actual result. To obtain a configurable zoom indicator, one would hence need to add a new variable with the corresponding content which is then populated by the renderer with the correct string (mag or Z or something like that). I am aware that this might be quite a lot to do for such a small change—but it seems as if this would be required to prevent a feature regression.
> 
> Regarding the tests, the situation is slightly messy. There is a `tests` folder, which contains `install.sh` and `test.sh` scripts which _should_ do what their names suggest.

Let me give a try based on this

PR #2193, opened 2022-01-20
Fix #2186

### LINKED ISSUE #2186: How do you insert static strings into powerline segments?
I have no idea why this isn't working:

```
{
  "segment_data": {
    "string": {
      "name": "foo",
      "contents": "bar"
    }
  },
  "segments": {
    "left": [
      {
        "string": "foo"
      }
    ]
  }
}
```

SIGSTACKFAULT:
Probably. Works On My Machine. tests/install.sh didn't work for me, but that should probably be a different issue.


SIGSTACKFAULT:
...and I just realized I typo'd the issue number.

SIGSTACKFAULT:
I can't be bothered to rename the branch

PR #2224, opened 2022-12-17
Feature: Check env vars for mail username and password

jessebot:
Hi powerline maintainers :)

This is so that you don't have to pass in username and password to the mail segment function, that way you can still check in your powerline config to source control and won't have to worry about leaking your password in plain text.

- Checks shell environment variables: `MAIL_USER`, `MAIL_PASSWORD`

I could also rework this so that the mail segment function takes optional variables called something like `mail_user_variable` and `mail_password_variable`, to support multiple segments with different accounts to check for mail.

Happy to accept feedback. :)


PH111P:
Hi @jessebot,
Thanks for your contribution, that is a great idea!

Indeed, I think it would be best if the environment variables would be configurable (but that should not be that much work to incorporate).

(BTW: I think that the base branch should be set to `develop`.)

PR #2225, opened 2022-12-18
Feature: add username_variable, password_variable, server_variable, port_variable to options to mail segment

jessebot:
This is a continuation of the proposal in #2224, but the base branch was `master` instead of `develop`, and I updated the feature branch name in the fork.

We also decided to add allowing users to specify the names of the shell environment variables they want to use, that way they can have different env variables for different segments.

Overall this adds 4 new parameters:

|  parameter              | description                |
|:-------------------:|:--------------------:|
| username_variable | string, name of env var to check for email username |
| password_variable | string, name of env var to check for email password |
| server_variable      | string, name of env var to check for email server    |
| port_variable         | string, name of env var to check for email server port    |

So a user could now create a segment like this:

```json
        {
            "function": "powerline.segments.common.mail.email_imap_alert",
            "args": {
                "username_variable": "MAIL_USER",
                "password_variable": "MAIL_PASS",
                "server_variable":   "MAIL_SERVER",
                "port_variable":     "MAIL_PORT",
                "use_ssl": false,
            }
        },
```

and then have those variables exported in BASH; example (this would let you use the protonmail-bridge):

```bash
export MAIL_USER="youruser@protonmail.com"
export MAIL_PASS="secret bridge password"
export MAIL_SERVER="127.0.0.1"
export MAIL_PORT=1143
```

### DIFF
--- a/powerline/segments/common/mail.py
+++ b/powerline/segments/common/mail.py
@@ -29,7 +29,7 @@ class EmailIMAPSegment(KwThreadedSegment):
 			password = os.environ[password_variable]
 		if server_variable:
 			server = os.environ[server_variable]
-		if port:
+		if port_variable:
 			port = os.environ[port_variable]
 
 		return _IMAPKey(username, password, server, port, folder, use_ssl)
--- a/powerline/segments/common/mail.py
+++ b/powerline/segments/common/mail.py
@@ -19,16 +19,16 @@ class EmailIMAPSegment(KwThreadedSegment):
 	interval = 60
 
 	@staticmethod
-	def key(username, password, server='imap.gmail.com', port=IMAP4_SSL_PORT, folder='INBOX', use_ssl=None, **kwargs):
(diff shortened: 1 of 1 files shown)


powerline/segments/common/mail.py
```diff
@@ -17,9 +19,19 @@ class EmailIMAPSegment(KwThreadedSegment):
...
 		if use_ssl is None:
 			use_ssl = (port == IMAP4_SSL_PORT)
+		# catch if user set custom mail credential env variables
+		if username_variable:
+			username = os.environ[username_variable]
+		if password_variable:
+			password = os.environ[password_variable]
+		if server_variable:
+			server = os.environ[server_variable]
+		if port:
```
PH111P: I think this should be `port_variable`.

PH111P:
Currently, there is no such thing as a release schedule (which also wouldn't make too much sense given the rather infrequent contributions these days). 
Having said that, perhaps it is time for a new release.

mehdi2021mp:
ec3c65b7ed20e0b563b13c6aadb1c7266c8ad42f

Issue #2116, opened 2020-08-14
Update powerline-status on pypi.org to 2.8.1


taisph:
https://pypi.org/project/powerline-status/ is still at 2.7. Could someone update it to the latest release?

PH111P:
I know and I would like to, but sadly I don't have access to pypi so I can't update the release there, sorry. (There is a similar issue with the documentation.)

In the meantime, you can install powerline via pip directly from the repo via
```
pip install git+https://github.com/powerline/powerline.git@develop
```

taisph:
> I know and I would like to, but sadly I don't have access to pypi so I can't update the release there, sorry. (There is a similar issue with the documentation.)

If the maintainers of the pypi package don't respond doesn't pypi.org support have some sort of procedure to change inactive maintainers by request? I doubt it is in anyone's interest to have packages bitrotting on pypi.org.


> In the meantime, you can install powerline via pip directly from the repo via
> ```
> pip install git+https://github.com/powerline/powerline.git@develop
> ```

Thanks. I actually didn't know pip could pull directly from GitHub.

taisph:
@graingert and @ZyX-I are listed as maintainers on pypi.org. If one of them could spare a few minutes to help you out and by extent powerline users, that would be fantastic.

graingert:
@taisph hello, if you make a PR for a GitHub action to upload releases to pypi I'll mint a token for it

StopMotionCuber:
@graingert just wanted to ping you as the PR is available at #2118

tigerinus:
No rush, but anytime soon will we be able to install latest version via pip?

Thanks.

PH111P:
> No rush, but anytime soon will we be able to install latest version via pip?
> 
> Thanks.

Please try `pip install git+https://github.com/powerline/powerline.git@develop`.

maciej-lech:
This should not be closed as 2.7 is still the latest version in pypi.

graingert:
Who has permissions to add secrets to this repository?

PH111P:
I think only @ZyX-I (seemingly inactive) and @Lokaltog can add secrets.

weavage:
What about this?
https://peps.python.org/pep-0541/#how-to-request-a-name-transfer

graingert:
You don't need a name transfer if I'm happy to add a secret to your GitHub actions

weavage:
Well shoot, what do we need to do to get the PyPi publishing going again?

[...]

PR #2255, opened 2024-04-06
Fix parsing of tmux version output

grom-42:
Hello!

I had an issue with the parsing of the command "tmux -V" which throw an error with my version of tmux, installed using "apt install tmux" on an Ubuntu 20.04.
Actually this version output "tmux next-3.4" and the parser did not handle the "next-" prefix.

So, here is my suggestion to fix it. :)

I used regexps only but, obviously, feel free to modify anything to suit your preferences. ;)

### DIFF
--- a/powerline/bindings/tmux/__init__.py
+++ b/powerline/bindings/tmux/__init__.py
@@ -67,19 +67,45 @@ def source_tmux_file(fname):
 	run_tmux_command('source', fname)
 
 
-NON_DIGITS = re.compile('[^0-9]+')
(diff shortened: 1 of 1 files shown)


schanzel:
This was fixed by https://github.com/powerline/powerline/pull/2033, but the version currently on PyPI (2.7) does not have the fix (cf. also this discussion https://github.com/powerline/powerline/issues/2116).
Installing the latest version from git should do the trick.
`pip install git+https://github.com/powerline/powerline.git@2.8.3`

PR #2266, opened 2024-10-04
Support aliases on cwd segments

christophehenry:
Add path_aliases argument to `CwdSegment.get_shortened_path` allowing to specify aliases for well-known directory paths.


christophehenry:
Uh, that's weird: this change doesn't seem to work when working in a Python venv. Actually `CwdSegment.__call__` doesn't seem event called.

PR #2271, opened 2025-01-10
Fix compatibility with Python 3.13 and up

sicherha:
* Fix a `TypeError` in the Vim statusline that occurs with Python 3.13.
* Fix a `ValueError` in `powerline-daemon` that will occur with Python 3.14 - nesting argument groups has been removed (https://bugzilla.redhat.com/show_bug.cgi?id=2336943).
* While at it, fix a bunch of typos.
* Fix the CI pipeline.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.11, 3.12]
+        python-version: [3.8, 3.9, 3.11, 3.12, 3.13]
 
     steps:
     - name: Checkout code
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, 3.11, 3.12, 3.13]
+        python-version: [3.8, 3.9, 3.11, 3.12, 3.13, 3.14]
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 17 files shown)


badshah400:
Hi, on openSUSE Tumbleweed, which just moved to python 3.13 as the default, we applied your patch to make the vim plugin work again. Many thanks. In addition, we also needed the following patch to fix a test file issue to get powerline building again with its test suite enabled.

```
Index: powerline-2.8.4/tests/test_python/test_cmdline.py
===================================================================
--- powerline-2.8.4.orig/tests/test_python/test_cmdline.py
+++ powerline-2.8.4/tests/test_python/test_cmdline.py
@@ -52,7 +52,7 @@ class TestParser(TestCase):
 			]:
 				self.assertRaises(SystemExit, parser.parse_args, raising_args)
 				self.assertFalse(out.getvalue())
-				self.assertRegexpMatches(err.getvalue(), raising_reg)
+				self.assertRegex(err.getvalue(), raising_reg)
 				flush()
 
 	def test_main_normal(self):
```

sicherha:
Thanks! I have added a commit to this PR that applies your diff.

And I'll see to it that Fedora's `powerline.spec` file, too, gets a `%check` section.

sicherha:
I have added a commit that fixes compatibility with Python 3.15. The fix requires at least Python 3.10, which I consider acceptable because Python 3.9 has been EOL since October 2025.

.github/workflows/main.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.11, 3.12]
+        python-version: [3.8, 3.9, 3.11, 3.12, 3.13]
```
cdayjr: does this need to be updated as well?

```suggestion
        python-version: [3.10, 3.11, 3.12, 3.13, 3.14]
```

sicherha: Good suggestion. Done in a separate commit.

sabedevops:
The patch on `powerline/commands/daemon.py` got it running for me on Fedora 43, thank you!

